PR #44, opened 2014-08-21
Add round-tripping to coordinate conversion

astrofrog:
This needs some more comments and tests

cc @Cadair


cdeil:
Is this still something that should go in?

astrofrog:
@cdeil - eventually I think so but I need to re-examine this once the sunpy coordinates module is finally merged, since that was the use case where it was needed (cc @Cadair)

Cadair:
*finally*.... eugh :p

astrofrog:
@Cadair - can you try out this branch? This includes the commits from https://github.com/astrofrog/reproject/pull/102 for now. I think it should fix your issue?

test

I need to add a regression test, and see if we can extend this to also work with the full n-dimensional reproject.

Cadair:
Awesome! I will lest later.

Cadair:
Seems to work ok (I think we ran the same code lol)

astrofrog:
Done in #129

PR #55, opened 2015-02-27
Add command line tool

cdeil:
How about adding a simple command line tool?

Something like:

```
$ astropy-reproject image.fits target.fits out.fits --type="flux-conserving"
```

@astrofrog – I could make a PR for this shortly. Let me know if you want such a command line tool and if you have any suggestions for the name and arguments.

Remaining TODOs:
- [ ] Add tests
- [ ] Add docs


cdeil:
I've written a first version (completely untested).

@astrofrog – What do you think about the command line options? (eventually I would like to use this for input HEALPIX images.) Which files should I use to test this?

reproject/scripts.py
```diff
@@ -0,0 +1,88 @@
...
+                  inext=0, targetext=0, overwrite=False,
+                  method='bilinear', store_footprint=True):
+    """
+    Reproject one image onto another (file-based interface).
+
+    File-based interface to `reproject.reproject`.
+
+    Parameters
+    ----------
+    TODO
```
cdeil: Do you want this function to be part of the public API (i.e. show up in the docs)?
If no, I could skip describing all the parameters here?

astrofrog: I think we don't usually have these in the API docs and instead we should have a section in the docs mentioning the command-line tool and maybe just showing the output of ``--help``.

astrofrog:
So as mentioned in https://github.com/astrofrog/reproject/issues/87, do we want to go ahead with something like this? In a way, we're opening ourselves up to maintaining a parallel command-line API, but maybe that's ok?

astrofrog:
Since this is not a commonly requested feature at the moment, and since the Python API has now grown to be perhaps more than we want to maintain with a command-line tool, I'm going to go ahead and close this, but if users request this in future we can consider opening it again. Thanks for your work on this @cdeil!

PR #60, opened 2015-04-25
Implement support for X-ray event lists

astrofrog:
@taldcroft has the following code:

https://github.com/taldcroft/xrayevents

It would be great to include this in this package, so that users can also call `reproject` in the same way as for other images but with a Table HDU or filename for the table as the first argument.

One thing I am not sure about is how to deal with the footprint.

@taldcroft, can we use your code here under a BSD license? Are you interested in trying to add support for this here? (if not, I can do it)


astrofrog:
@taldcroft - implemented! Would you mind reviewing this and trying it out briefly to make sure I'm not doing anything silly? Note that this is less convenient than your script for the case where the user just wants to bin assuming the existing pointing in the events list, but is more consistent with the API in the rest of the reproject package (the advantage is when one needs to reproject event lists and normal images to the same header).

cdeil:
Hmm ... I have something very similar in `gammapy.data.EventList.fill_counts_header`:
```
"""Fill events in counts image specified by a FITS header.
```
https://github.com/gammapy/gammapy/blob/master/gammapy/data/event_list.py#L269

It doesn't do weights and my x, y columns are named differently, but it would be very easy to change / extend that to work for x-ray data as well.

Also , there may or may not be a student soon that implements the `histogram` functionality for `NDData`
https://github.com/astropy/astropy/issues/619

I'm not sure combining `pix_to_world` and `histogram` in a `reproject_events` function in this package is the right way / place to put this functionality, I always assumed `reproject` was only about reprojecting pixels only. But OK, fine with me if you want this here, @astrofrog.

@taldcroft, @astrofrog  – the `event_filter` function is nice:
https://github.com/taldcroft/xrayevents/blob/master/xrayevents.py#L50
Would it be in scope as a method on `Table`, maybe even with some string parsing?
```python
table = table.filter('(42 < energy <= 43.5) & (0 < dec <  53)')
```

taldcroft:
@astrofrog - OK I'll have a look.

@cdeil - the filter idea with string parsing looks good to me.  PyTables has a method `read_where` that does something similar so there might be some parsing code there that could be re-used.  This might be something for @mdmueller and @mdboom to consider.

astrofrog:
I think it would be nice to have this kind of functionality here because it could also be used for e.g. binning star counts, etc. I could rename it to something more generic.

As an aside, I wonder if there is room for a Table subclass for event lists, which would have a mix-in ``Time`` column, a mix-in ``SkyCoord`` column, and other optional arrays? Then the filtering would be easy using the normal table indexing.

cdeil:
I've implemented `time`, `energy`, `radec` and a few other things as attributes in the EventList class:
https://gammapy.readthedocs.org/en/latest/api/gammapy.data.EventList.html
There's also `select_time`, `select_energy`, `select_sky_cone` and `select_sky_box` methods.

Any feedback on how to implement this or any other class in Gammapy would be very welcome.
E.g. if you think mix-ins for `time`, `radec`, ... is better for some reason, I'm happy to change.

astrofrog:
Ok, so coming back to this, I think there are two distinct things here:

* I would like to have a function like that one here that simply takes an array ``SkyCoord`` and some weights and then bins it onto a header, because that is generically useful, e.g. for event-based data but also for catalogs, etc.

* I don't want to reimplement a whole event file parsing framework including the ability to filter, etc. since it's a bit beyond the scope here. However, we should think about whether we can combine what exists in gammapy for event files with functionality that would also work for x-ray event files. I don't see why we couldn't move the gammapy EventList functionality to the core package since it's not really gamma-ray specific, right? (though I'm not sure where all the file format subtleties of gamma-ray vs X-ray would fit in best). I could then support accepting this eventlist class and extracting data from it, but I wouldn't have to reimplement all the I/O logic. @cdeil @taldcroft - any thoughts on this?

cdeil:
@astrofrog I agree the `SkyCoord` histogram functionality is generally useful. But shouldn't this be added as an `NDData` method or utility function in `astropy.nddata.utils`?

A long time ago I proposed this in https://github.com/astropy/astropy/issues/619 and now implementing this is planned for week 1 of @mapazarr's GSoC and we wanted to talk about this tomorrow. Any feedback on the proposed functionality / API in that issue are very welcome ... especially @astrofrog, if you now think this should be implemented as a function in `reproject` instead of in `astropy.nddata`.

Concerning the Gammapy `EventList` class, it can't be moved to the Astropy core. There's an [OGIP standard from the 90s for X-ray astronomy](https://heasarc.gsfc.nasa.gov/docs/heasarc/ofwg/docs/events/ogip_94_003/ogip_94_003.html), but yes, there will be some gamma-ray specific additions ... the spec for CTA is being written, it probably will not be final for a few more years, i.e. is not stable at all and will require code changes from time to time.

What is the motivation for this PR? Was there a use case to be able to create counts images from event lists like e.g. ds9 does it?

astrofrog:
@cdeil - I'd be perfectly happy for this to be implemented in the core instead, so by all means, please do feel free to implement it in ``astropy.nddata``.

The motivation for this PR is that indeed, like ds9, it would be cool to be able to read in a FITS file seamlessly even if it's stored as events on disk. But I'd be happy to point users to the core function if it's just as easy to use.

astrofrog:
Note that the ``convert_world_coordinates`` function in ``reproject`` will be needed in the core if you want to support rebinning onto arbitrary WCS. I think it should go in there anyway since it's a pretty useful function, so I'll open a pull request soon.

astrofrog:
I'm going to close this as I think looking back at this it is maybe indeed beyond the scope of this package.

PR #65, opened 2015-04-30
Refactor HEALPIX coordinate conversion

astrofrog:
This updates the HEALPIX projection code to use Astropy coordinate conversions which is much more robust to differences in e.g. equinox, and supports more coordinate systems. On the downside, we lose support for Ecliptic coordinate conversion but that will be temporary since I think @eteq is going to be implementing this soon in Astropy :smile:

Another thing that has changed here is the API to the functions, to bring it in line with other parts of the reproject package.

This is a WIP because I can't install healpy at the moment (https://github.com/healpy/healpy/issues/248) so I have to test this via Travis.


cdeil:
@astrofrog – Just in case you have Macports and want to test locally: `port install py34-healpy` works.

astrofrog:
@cdeil - thanks, that worked!

astrofrog:
@lpsinger @cdeil @keflavich - this should now be ready for testing/review. Note the slightly changed function calls for ``healpix_to_image`` and ``image_to_healpix``. This does a much better job at coordinate conversion excluding the ecliptic conversion.

We still need to expose this via the high-level ``reproject`` function and write some docs, but this will do for now for this pull request.

astrofrog:
Ah, I need to update the docstrings too...

astrofrog:
I'm going to go ahead and merge this, but if anyone using this runs into any problems, please open an issue!

PR #67, opened 2015-05-04
Refactor API and docs

astrofrog:
This is a large refactor that does the following:

* Complete re-write of the docs
* Expose functions for different algorithms (e.g. ``reproject_interpolation``, etc.)
* New high-level HEALPIX functions

I'm actually not sure if we need the 'high-level' ``reproject`` anymore now because it doesn't really add much and makes it harder to pass arguments to the sub-functions. I'm considering removing it for now, since the algorithm-specific functions are now still very high-level and can take a variety of inputs.

This also exposes the HEALPIX functionality at the top-level.

**Preview of docs: http://www2.mpia-hd.mpg.de/~robitaille/astropy/reproject_docs/**


cdeil:
Thanks!!!
The diff is a bit hard to read, can you maybe post the html docs online?

astrofrog:
@cdeil - here you go:

http://www2.mpia-hd.mpg.de/~robitaille/astropy/reproject_docs/

It's lacking in figures at the moment, but I'll leave that for a future PR since this one is already significant...

docs/celestial.rst
```diff
@@ -0,0 +1,132 @@
...
+  <http://montage.ipac.caltech.edu/index.html>`_ package, and we have
+  implemented it here using the same core algorithm.
+
+Currently, this package implements interpolation and spherical polygon
+intersection.
+
+The reprojection/resampling is always done assuming that the image is in
+**surface brightness units**. For example, if you have an image with a constant
+value of 1, reprojecting the image to an image with twice as high resolution
+will result in an image where all pixels are all 1.
```
cdeil: What if someone wants to measure flux (not surface brightness) on the reprojected image?
They need a utility function or at least advice to do this, no?

astrofrog: Just to check, do you mean that they want to preserve flux? Or measure a flux for a source?

docs/celestial.rst
```diff
@@ -0,0 +1,132 @@
...
+celestial. There are several existing algorithms that can be used to
+reproject such data:
+
+* **Interpolation** (such as nearest-neighbor, bilinear, biquadratic
+  interpolation and so on). This is the fastest algorithm and is suited to
+  common use cases, but it is important to note that it is not flux
+  conserving, and will not return optimal results if the input and output
+  pixel sizes are very different.
+
+* **Drizzling**, which consists of determining the exact overlap fraction of
```
cdeil: Is there a good reference to a page or paper describing drizzle?

astrofrog: http://arxiv.org/abs/astro-ph/9808087 - will add to these docs

docs/healpix.rst
```diff
@@ -0,0 +1,55 @@
+**************
+HEALPIX images
```
cdeil: The `healpix.rst` and HEALPIX high-level docstrings don't really say what method is used for reprojection.
I think it's bilinear interpolation in both directions?
Can you state this?

astrofrog: Yes, I should clarify, and I think we could allow more fine-tuned control over the order of the interpolation.

reproject/healpix/utils.py
```diff
@@ -0,0 +1,53 @@
...
+                return system_new
+
+
+def parse_input_healpix_data(input_data, field=0, hdu_in=None):
+    """
+    Parse input HEALPIX data to return a Numpy array and coordinate frame object.
+    """
+
+    if isinstance(input_data, (TableHDU, BinTableHDU)):
+        # TODO: for now we have to write out to a temporary file
```
cdeil: What do you mean by "for now"?
Is there a feature request in healpy or Astropy to implement something that's needed here?
Maybe leave a reference to it here or as a `reproject` Github issue?

docs/celestial.rst
cdeil: Say you have a star or Galaxy on a calibrated flux image on which you could measure the flux using aperture or PSF-fitting photometry. Now you reproject the image. How do you get the flux from the reprojected image?

(I don't know if that's a common use case for optical astronomers ... I do have a use case for this ... measure flux of sources on reprojected and stacked images.)

reproject/healpix/high_level.py
```diff
@@ -0,0 +1,90 @@
...
+            * An image HDU object such as a `~astropy.io.fits.PrimaryHDU` or
+              `~astropy.io.fits.ImageHDU` instance
+            * A tuple where the first element is a `~numpy.ndarray` and the
+              second element is either a `~astropy.wcs.WCS` or a
+              `~astropy.io.fits.Header` object
+
+    coord_system_out : `~astropy.coordinates.BaseCoordinateFrame` or str
+        The output coordinate system for the HEALPIX projection
+    hdu_in : int or str, optional
+        If ``input_data`` is a FITS file or an ~astropy.io.fits.HDUList`
```
cdeil: Missing backtick (see [here](http://www2.mpia-hd.mpg.de/~robitaille/astropy/reproject_docs/api/reproject.reproject_to_healpix.html))

reproject/spherical_intersect/high_level.py
```diff
@@ -0,0 +1,61 @@
...
+              `~astropy.io.fits.Header` object
+
+    output_projection : `~astropy.wcs.WCS` or `~astropy.io.fits.Header`
+        The output projection, which can be either a `~astropy.wcs.WCS`
+        or a `~astropy.io.fits.Header` instance.
+    shape_out : tuple, optional
+        If ``output_projection`` is a `~astropy.wcs.WCS` instance, the
+        shape of the output data should be specified separately.
+    hdu_in : int or str, optional
+        If ``input_data`` is a FITS file or an ~astropy.io.fits.HDUList`
```
cdeil: Missing backtick.

reproject/interpolation/high_level.py
```diff
@@ -0,0 +1,78 @@
...
+              `~astropy.io.fits.Header` object
+
+    output_projection : `~astropy.wcs.WCS` or `~astropy.io.fits.Header`
+        The output projection, which can be either a `~astropy.wcs.WCS`
+        or a `~astropy.io.fits.Header` instance.
+    shape_out : tuple, optional
+        If ``output_projection`` is a `~astropy.wcs.WCS` instance, the
+        shape of the output data should be specified separately.
+    hdu_in : int or str, optional
+        If ``input_data`` is a FITS file or an ~astropy.io.fits.HDUList`
```
cdeil: Missing backtick.

docs/celestial.rst
cdeil: I think this is important enough to highlight in an `.. info::` box.

docs/healpix.rst
cdeil: Could you use an example data file in `healpix.rst` that the user can get?

The descriptions are good, but it would be even much nicer if I could actually copy & paste the code into an IPython notebook and start playing with this ...
See http://www2.mpia-hd.mpg.de/~robitaille/astropy/reproject_docs/healpix.html

cdeil:
:+1: I really like how the different methods can be executed via a uniform API!

Like you said, I'm also unsure whether a `reproject` function that wraps the different methods is a good idea. On the one hand it can be useful to say
```
for method in ['interp', 'exact']:
     reproject(..., method=method)
```
in a script, on the other hand more functions are bad and parameters that only apply to some methods can be confusing ...

docs/index.rst
```diff
@@ -1,10 +1,76 @@
...
+*******************************
+
+Introduction
+============
+
+The *reproject* package implements image reprojection (resampling) methods
+for astronomical images and more generally n-dimensional data. These assume
+that the WCS information contained in the data are correct. This package does
+**not** do image registration, which is the process of aligning images where
+one or more image may have incorrect or missing WCS.
```
cdeil: typo: "image" -> "images"

reproject/healpix/high_level.py
```diff
@@ -0,0 +1,90 @@
...
+        If ``input_data`` is a FITS file, specifies the HDU to use.
+
+    Returns
+    -------
+    array_new : `~numpy.ndarray`
+        The reprojected array
+    footprint : `~numpy.ndarray`
+        Footprint of the input array in the output array. Values of 0 indicate
+        no coverage or valid values in the input image, while values of 1
+        indicate valid values. Intermediate values indicate partial coverage.
```
cdeil: "Intermediate values indicate partial coverage." -> I think you're only using integer 0 or 1 here, no intermediate values?

Generally a description and example of what exactly footprint is and what happens at the border of the input image in the Sphinx docs would be nice for users that care about this.

reproject/healpix/utils.py
astrofrog: I did a pull request to healpy recently to allow it to support HDU and HDUList input but we have to wait for a stable release before supporting that (see https://github.com/healpy/healpy/pull/249). I added a note here.

docs/celestial.rst
astrofrog: Done

reproject/healpix/high_level.py
astrofrog: Fixed

reproject/spherical_intersect/high_level.py
astrofrog: Fixed

reproject/interpolation/high_level.py
astrofrog: Fixed

docs/index.rst
astrofrog: Fixed

reproject/healpix/high_level.py
astrofrog: Good catch, and I agree, a more detailed description of footprints would be good.

astrofrog:
Just summarizing below the things I need to do:

* [ ] Better explain the conservation of surface brightness and how it impacts photometry, and how one can preserve e.g. counts instead.
* [ ] Include example files the users can get to reproduce the examples
* [ ] Better explain the footprint

@cdeil - regarding looping over methods, that was the initial motivation for the ``reproject`` function but then I realized that in practice that use case is rare. People are likely going to choose one method and stick to it, with the exception of when they are testing how the different algorithms differ.

cdeil:
Actually, with the current package, is it possible to reproject and preserve counts (or equivalently flux)?

I think no, one would need either a weights image how many fractional input pixels correspond to each output pixel, or pixel solid angle images?

(Here's some work in progress code to compute solid angle images –  https://github.com/gammapy/gammapy/pull/102, but this doesn't belong in Gammapy, it should be in `reproject` or Astropy core.)

astrofrog:
@cdeil - this is indeed not possible at the moment and does require a function that would give the pixel area for each pixel (related to https://github.com/astrofrog/reproject/issues/6)

cdeil:
OK – So implementing flux-conserving reprojection should probably be postponed to a separate PR because this one is already big.

astrofrog:
@cdeil - indeed, I don't think I can address it here, but I can add a note.

astrofrog:
@cdeil - I just spent all this evening working on the docs and adding plots in various places. The main changes are:

* Expanded 'getting started' section with plots
* A new section of the docs on footprint arrays with plots to demonstrate the different types
* Expanded HEALPIX docs with a plot (note that the plot has an issue with a black frame but this is a WCSAxes issue I need to fix, nothing I can do here for now)
* Made a proper note about surface brightness vs flux in the celestial section

The preview is here: http://www2.mpia-hd.mpg.de/~robitaille/astropy/reproject_docs/

reproject/healpix/high_level.py
astrofrog: Done

docs/healpix.rst
astrofrog: Done

docs/celestial.rst
astrofrog: Done

reproject/healpix/high_level.py
cdeil: This is still unchanged, no?
So does healpix now ever return footprint values other than the integer 0, 1?

docs/healpix.rst
```diff
@@ -0,0 +1,167 @@
...
+    CUNIT1  = 'deg     '
+    CTYPE2  = 'DEC--MOL'
+    CRPIX2  =                  400
+    CRVAL2  =                  0.0
+    CDELT2  =                  0.4
+    CUNIT2  = 'deg     '
+    COORDSYS= 'icrs    '
+    """, sep='\n')
+
+All of the following are examples of valid ways of reprojecting the HEALPIX LIGO data onto the Aitoff projection:
```
cdeil: "Aitoff" -> "Mollweide"?

astrofrog:
Ok, I believe this is now done :sweat: 

@cdeil - let me know if you have any other comments, and otherwise I will merge provided Travis passes.

docs/healpix.rst
astrofrog: Fixed

cdeil:
It's still not clear to me from the docs what kind of interpolation is performed for HEALPIX:
http://www2.mpia-hd.mpg.de/~robitaille/astropy/reproject_docs/healpix.html
http://www2.mpia-hd.mpg.de/~robitaille/astropy/reproject_docs/api/reproject.reproject_from_healpix.html

About HEALPIX: pixels are 4-point polygons with edges that are great circles, no?
So it should be possible to re-use the Montage code to get an "exact" reprojection and thus flux-preserving reprojection in the future for HEALPIX also?
@astrofrog If this is the case, maybe it's worth considering that possible extension for the API here or make a new issue to take a not to think about that later?

reproject/healpix/high_level.py
astrofrog: Huh I thought I had fixed that but apparently not. Done now. The answer to your question is no, it does not return anything other than 0,1 for healpix.

cdeil:
@astrofrog – This is awesome, thanks!!!

I'll start using the HEALPIX reprojection functions for my applications as soon as this is in master...

astrofrog:
@cdeil - I've now renamed the functions for HEALPIX to include 'interp' in the name and added a sentence to the documentation. I agree that one could indeed do the exact method in future in principle.

astrofrog:
The names are getting a bit long though... :-/ If you have any better suggestions, let me know! Otherwise we can go with these for now.

astrofrog:
@cdeil - maybe ``interp`` isn't needed in the HEALPIX function names. I don't know if anyone's ever going to have the time to implement other modes, but even if they do we can always then rename it to _interp and keep the original one for backward compatibility, right? Since you'll actually be using these on a daily basis, let me know what you think :)

cdeil:
Agreed it doesn't help to add `_interp` to the HEALPIX function names now.

The travis-ci docs build failed ... is it this warning that's the cause?
```
$ python setup.py $SETUP_CMD
reproject/spherical_intersect/setup_package.py:1: RuntimeWarning: Parent module 'reproject.spherical_intersect' not found while handling absolute import
  import os
reproject/spherical_intersect/setup_package.py:2: RuntimeWarning: Parent module 'reproject.spherical_intersect' not found while handling absolute import
```
https://travis-ci.org/astrofrog/reproject/jobs/61423659#L849

Otherwise :+1: to merge.

I'll make a new issue with the feature request for the "exact" reproject mode with HEALPIX.
For my current applications the interpolation mode is OK, but if someone wants to do photometry and convolutions in HEALPIX and compare to normal images, an "exact" flux-preserving reproject function would be helpful.

astrofrog:
@cdeil - I don't think that those warnings are what caused the build to fail, but instead I think it's due to the build terminating early. However, the error message isn't visible, so re-running without the ``-w`` option.

astrofrog:
Travis is a little flakey with the docs build so will look into it after this is merged in. Will also address the decreased coverage in a future PR, but for now this is fine.

astrofrog:
@cdeil - thanks for the reviews! Let me know if you run into any issues in real-life testing

PR #69, opened 2015-05-06
Add travis-ci build without optional dependencies

astrofrog:
@astrofrog I saw that you used `@pytest.mark.importorskip('healpy')` here
https://github.com/astrofrog/reproject/blob/dd3086eab4802749d8cd70a55ca920d70a6e0d77/reproject/healpix/tests/test_healpix.py#L40
and wanted to use this in Gammapy, but it doesn't work:
https://travis-ci.org/gammapy/gammapy/jobs/61290205#L907
https://github.com/gammapy/gammapy/pull/258

This also doesn't work for me locally ... is this the correct way to use this mark?

Maybe the `reproject` tests on travis-ci don't fail just because there's no build where healpy is not available:
https://github.com/astrofrog/reproject/blob/master/.travis.yml#L96
I guess that should be added.


astrofrog:
There are fixes for this in https://github.com/astrofrog/reproject/pull/77. Will rebase once that's merged.

PR #61, opened 2015-05-06
Add support for filename in first argument

astrofrog:
A common use case will be to pass the filename to a FITS or ASDF file as the first argument.


astrofrog:
This already worked but was not documented

PR #78, opened 2015-05-06
Ignore Numpy warnings in HEALPIX

astrofrog:


astrofrog:
@cdeil - I believe the warnings you were seeing should be gone after this

Issue #76, opened 2015-05-06
reproject_exact crashes sometimes with Macports


cdeil:
I was getting a crash executing the example from the docs [here](http://reproject.readthedocs.org/en/latest/footprints.html):
```python
from astropy.io import fits
from astropy.utils.data import get_pkg_data_filename
hdu1 = fits.open(get_pkg_data_filename('galactic_center/gc_2mass_k.fits'))[0]
hdu2 = fits.open(get_pkg_data_filename('galactic_center/gc_msx_e.fits'))[0]
from reproject import reproject_interp, reproject_exact
array_interp, footprint_interp = reproject_interp(hdu2, hdu1.header)
array_exact, footprint_exact = reproject_exact(hdu2, hdu1.header)
```
The crash report is here: https://gist.github.com/cdeil/cda42d1c5357e053b1c2

This is with Python 3.4 from Macports and from an IPython  notebook.
I don't know what the problem is.
The IPython console output mentions a CFITSIO version mismatch:
```
$ ipython notebook
[I 22:10:11.655 NotebookApp] Using MathJax from CDN: https://cdn.mathjax.org/mathjax/latest/MathJax.js
[W 22:10:11.699 NotebookApp] Terminals not available (error was No module named 'terminado')
[I 22:10:11.699 NotebookApp] Serving notebooks from local directory: /private/tmp
[I 22:10:11.699 NotebookApp] 0 active kernels 
[I 22:10:11.699 NotebookApp] The IPython Notebook is running at: http://localhost:8888/
[I 22:10:11.699 NotebookApp] Use Control-C to stop this server and shut down all kernels (twice to skip confirmation).
[I 22:10:29.713 NotebookApp] Creating new notebook in 
[I 22:10:30.557 NotebookApp] Kernel started: 1e1521ba-8ed9-47a1-ad7c-dcc47599478e
[I 22:11:42.558 NotebookApp] KernelRestarter: restarting kernel (1/5)
WARNING:root:kernel 1e1521ba-8ed9-47a1-ad7c-dcc47599478e restarted
[I 22:12:31.523 NotebookApp] Saving file at /Untitled.ipynb
[I 22:14:31.521 NotebookApp] Saving file at /Untitled.ipynb
[I 22:16:31.518 NotebookApp] Saving file at /Untitled.ipynb
[I 22:18:32.513 NotebookApp] Saving file at /Untitled.ipynb

[...]

PR #80, opened 2015-05-07
Simplify the handling of the parallel processing in exact mode

astrofrog:
*may* solve #76

### DIFF
--- a/reproject/spherical_intersect/core.py
+++ b/reproject/spherical_intersect/core.py
@@ -169,6 +169,8 @@ def _reproject_celestial(array, wcs_in, wcs_out, shape_out, parallel=True, _lega
 
         results = pool.map(_reproject_slice, inputs)
 
+        pool.close()
+
         array_new, weights = zip(*results)
 
         array_new = sum(array_new)
--- a/reproject/spherical_intersect/core.py
+++ b/reproject/spherical_intersect/core.py
@@ -3,23 +3,27 @@ from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
 import signal
-import warnings
 
 import numpy as np
-
-from astropy.utils.exceptions import AstropyUserWarning
 
 from ..wcs_utils import convert_world_coordinates
 
 from ._overlap import _compute_overlap
 
 
-# Function to disable ctrl+c in the worker processes.
 def _init_worker():
+    """
+    Function to disable ctrl+c in the worker processes.
+    """
     signal.signal(signal.SIGINT, signal.SIG_IGN)
 
 
-def _reproject_celestial(array, wcs_in, wcs_out, shape_out, parallel=True, _method="c"):
+def _reproject_slice(args):
+    from ._overlap import _reproject_slice_cython
+    return _reproject_slice_cython(*args)
+
+
+def _reproject_celestial(array, wcs_in, wcs_out, shape_out, parallel=True, _legacy=False):
 
     # Check the parallel flag.
     if type(parallel) != bool and type(parallel) != int:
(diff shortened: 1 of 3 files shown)


astrofrog:
@cdeil - looks like Travis is happy so I will plan to merge this anyway, but curious whether it solves your issue?

cdeil:
@astrofrog – I just got back to this ... still testing.

I've managed to boil down the testcase to this:
https://gist.github.com/cdeil/709802defe5fc9f50177

Put this in an IPython notebook cell. Execute it once – OK. Execute it again - Crash.
Remove the `imshow` line – never crashes.

This only happens in the IPython notebook ... it must have to do with IPython and / or reproject making assumptions about the input data array that are invalid at the C level. My guess this is a bug in reproject.
I'll test this branch now.

You still can't reproduce with reproject master?

cdeil:
I'm getting the same crash with this branch.
@astrofrog – Do you want me to continue debugging on this branch or master?

Any better idea than adding print statements to the reproject_exact code?

astrofrog:
@cdeil - could you debug on this branch? The code is cleaner so I prefer it anyway. Thanks!

astrofrog:
@cdeil - by the way, what version of IPython are you using? (just trying to also reproduce that here and failing).

cdeil:
I'm using `py34-ipython @3.1.0_0` and `py34-matplotlib @1.4.3_0+cairo+pyside+webagg` from Macports on OS X 10.10 and Astropy and reproject master (or this branch).

cdeil:
PS: I don't think this should stop the 0.1 release. Crashes are rare and scary for Python users, but probably it's unlikely that this will affect many users?

astrofrog:
I can't reproduce the crash with the same versions but with conda. I will try MacPorts later. But it would be great if you could narrow down at what point in the code the crash happens. Can you check that this does not happen with ``parallel=False`` for ``reproject_exact``?

cdeil:
I've never seen the crash with `parallel=False`.

I only get it from an IPython notebook with two calls to `reproject_exact` in different cells. If I move the second call in the first cell it doesn't crash.

I guess that's a strong hint that the issue has to do with parallel execution, but I'm not sure.

reproject/spherical_intersect/core.py
```diff
@@ -130,69 +134,49 @@ def _reproject_celestial(array, wcs_in, wcs_out, shape_out, parallel=True, _meth
...
-            raise
-
-        finally:
-            if not pool is None:
-                # Clean up the pool, if still alive.
-                pool.close()
-                pool.join()
+        # Prime each process in the pool with a small function that disables
+        # the ctrl+c signal in the child process.
+        pool = Pool(nproc, _init_worker)
```
astrofrog: @cdeil - can you put print statements around here to see if this is where it's happening? This is where the processes are launched.

cdeil: I did put print statements.
The crash happens a few lines below: the `pool.close()` call in line 172.

astrofrog:
@cdeil - by the time the processes are closed all the reproject code should be done, so this is very weird. Can you try running just:

```
def identity(x):
    return x
    
def do():
    
    import multiprocessing as mp
    
    pool = mp.Pool(processes=10)
    
    pool.map(identity, range(10))
    
    pool.close()
    
do()
do()
```

?

cdeil:
Your simple example with `do()` never crashes.

astrofrog:
@cdeil - just one quick favor, can you confirm that the following crashes for you?

http://nbviewer.ipython.org/gist/astrofrog/89e4c4b4889e8e412f56

cdeil:
Yes, this does crash for me.
https://gist.github.com/cdeil/decb454a07ac871bf4d5

Do you understand this?
Where to file an issue about this?

astrofrog:
See https://github.com/astrofrog/reproject/issues/76#issuecomment-99935672 - since happens only in the notebook, I'd say it's a notebook bug.

astrofrog:
Also, I have NO IDEA what this is all about. But I'm happy that we have a MWE and it's not a reproject issue

astrofrog:
In any case, I think this is good to go - it's an improvement over the existing code as it's simpler.

PR #96, opened 2016-03-26
Generalize reprojection in 3 dimensions

keflavich:
Work was collaborative with @adonath

### DIFF
--- a/reproject/interpolation/core.py
+++ b/reproject/interpolation/core.py
@@ -2,6 +2,7 @@
 from __future__ import absolute_import, division, print_function
 
 import numpy as np
+from astropy import wcs
 
 from ..wcs_utils import convert_world_coordinates
 from ..array_utils import iterate_over_celestial_slices, pad_edge_1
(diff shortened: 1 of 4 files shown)


astrofrog:
Things that need to be done before merging:

* [ ] Add checks of units etc for the full reproject
* [ ] Add unit tests for full reproject
* [ ] Fix sphinx failure

keflavich:
Added 1x test for `_full_reproject` for the case where it matches `_reproject_celestial` on a cube.  Still needs more...

keflavich:
I think the test coverage should now be pretty good.  There are tests that check that `_reproject_full` and `_reproject_celestial` match when they should

keflavich:
failure is:
```
>       p_out = np.meshgrid(*p_out_ax, indexing='ij')
E       TypeError: meshgrid() takes exactly 2 arguments (4 given)
```
Suggest dropping np1.6 support, raising an Exception at this point if np<1.7 is used and putting an xfail in the tests.

astrofrog:
Yeah, let's drop Numpy 2.6

astrofrog:
@keflavich - at the moment, this will not do the right thing for spectral cubes that have velocity axes that use different conventions, correct? So I had an idea - what about we re-use some of the spectral-cube machinery (importing from spectral-cube, not copying) so first make sure we have the two WCS in the same exact conventions?

astrofrog:
Hmm actually maybe what's being done here is fine since it checks CTYPE.

astrofrog:
I'll go ahead and merge this now and we can refine in future PRs. Thanks @keflavich and @adonath!

keflavich:
Yes I want to use the spectral cube machinery for this.   It's done in the
associated spectral cube pr.  Eventually can merge that functionality in
here.

PR #86, opened 2016-04-04
Read other columns of HEALPIX image?

astrofrog:
Great package!  Would be great to be able to specify which column of a HEALPIX file to read.  This appears to be the "field" parameter in parse_input_healpix_data.  For the Planck foreground component images (I'm interested in the CO) the additional columns can be important.

If you do release an update, do I just repeat the "pip install --user reproject.py" (I'm using macports).  Or is there an update command?

Thanks,
Tony

### DIFF
--- a/reproject/healpix/high_level.py
+++ b/reproject/healpix/high_level.py
@@ -5,7 +5,7 @@ from ..utils import parse_input_data, parse_output_projection
 __all__ = ['reproject_from_healpix', 'reproject_to_healpix']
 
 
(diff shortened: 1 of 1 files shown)


astrofrog:
@tonywong94 - to install an update, just do

```python
pip install reproject --user --upgrade --no-deps
```

and this will update just the reproject package to the latest version.

Regarding your first question, I've attached code to this issue to do exactly this. Once this is merged, you will be able to pass the field= option to ``reproject_from_healpix``

PR #105, opened 2016-06-29
Significant improvements to performance for interpolation

astrofrog:
When dealing with large input images, we now make sure that we never force the whole image to be read (when memory-mapped) if only a small region is needed.

@pllim - this should fix the memory issues you were seeing. Can you install reproject from my branch to check if it works correctly, and once we've confirmed this I can go ahead and merge (if tests pass)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 - Allow users to pass a ``field=`` option to ``reproject_from_healpix``
   to access different fields in a HEALPIX file. [#86]
 
+- Significant improvements to performance when the input data is a large
+  memory-mapped array. [#105]
+
 0.2 (2015-10-29)
 ----------------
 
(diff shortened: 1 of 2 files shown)


pllim:
@astrofrog , thanks! Doesn't look like this fixed my use case. Still having `MemoryError`. Does it matter if I have `reproject` 0.2 from `pip install`, and then overwrote it (without deleting 0.2) by `python setup.py install` from your branch? Python seems to be picking up the latter:

```
>>> import reproject
>>> reproject.__version__
'0.3.dev321'
```

astrofrog:
@pllim - ok, I'll look into it with the specific mosaic you are using.

pllim:
I could also be that I'm giving it the wrong reprojection WCS. I can't tell what is causing this... :(

PR #124, opened 2017-03-23
Fix regression that sent some all-sky images to all NaNs

lpsinger:
This fixes a regression that was introduced in reproject-0.3 by commit 59c5b493c7807f70b19f0e337a88a9e0f1a3eccf that caused certain all-sky images (e.g. the Mellinger Milky Way Panorama, http://www.milkywaysky.com) to be reprojected to all NaNs when the output WCS was in Mollweide coordinates.

The regression was due to a change that converted the minima and maxima of reprojected indices from floats to integers so that Numpy would not complain about non-integer array indices. If any of the reprojected indices were NaN, then the minima and maxima would be NaN. Converting the indices to integers broke some all- sky reprojections because some subsequent code was inadvertently relying on the fact that comparisons between NaNs and other floating point values are always false.

The workaround is to find the minima and maxima of the non-NaN values of the reprojected indices using `np.nanmin` and `np.nanmax` respectively.

I added one unit test that uses the FITS header from the Mellinger data file, but a fake test image that is all ones.

## Test script

Here's a test script that demonstrates the issue:

```python
#!/usr/bin/env python
from astropy.utils.data import download_file
from astropy.io import fits
from astropy.stats import sigma_clipped_stats
from astropy.wcs import WCS
from astropy.visualization.wcsaxes.frame import EllipticalFrame
from reproject import reproject_interp
from matplotlib import pyplot as plt
import numpy as np

# Download and read Mellinger Milky Way panorama
url = 'http://galaxy.phy.cmich.edu/~axel/mwpan2/mwpan2_RGB_3600.fits'
filename = download_file(url, cache=True)
hdu, = fits.open(filename)
wcs = WCS(hdu.header, naxis=2)

# Measure image statistics
image_stats = np.asarray([sigma_clipped_stats(_) for _ in hdu.data])
mean = image_stats[:, 0].reshape(-1, 1, 1)
median = image_stats[:, 1].reshape(-1, 1, 1)
stdev = image_stats[:, 2].reshape(-1, 1, 1)

# Normalize image data for plotting with Matplotlib
normed_data = np.clip((hdu.data - mean) / stdev * 0.05 + 0.1, 0, 1)

degperpix = 0.25
plot_header = fits.Header(dict(
    NAXIS=2,
    NAXIS1=int(360 // degperpix), NAXIS2=int(180 // degperpix),
    CRPIX1=180 / degperpix, CRPIX2=90 / degperpix,
    CRVAL1=0, CRVAL2=0,
    CDELT1=-2*np.sqrt(2)/np.pi * degperpix,
    CDELT2=2*np.sqrt(2)/np.pi * degperpix,
    CTYPE1='GLON-MOL',
    CTYPE2='GLAT-MOL',
    RADESYS='ICRS').items())
plot_wcs = WCS(plot_header)

img = np.transpose([
    reproject_interp((_, wcs), plot_header)[0] for _ in normed_data], [1, 2, 0])

fig = plt.figure(figsize=(8, 5))
ax = plt.axes(projection=plot_wcs, frame_class=EllipticalFrame)
ax.imshow(img, origin='lower')
fig.savefig('test.png')
```

## Bad output

Here is the bad output, before this patch:

test-bad

## Good output

Here is the good output, either before the regression was introduced or after this patch is applied:

test-good


astrofrog:
@lpsinger - thanks for the fix! The Travis failures are unrelated, so I'll go ahead and merge.

lpsinger:
Would you please issue a new release? I'd like to use this function in some production code that is affected by the regression.

lpsinger:
> Would you please issue a new release? I'd like to use this function in some production code that is affected by the regression.

Ping!

lpsinger:
Would it be possible to issue a point release to cherry-pick this regression fix? I am using the master branch in development branches of LIGO/Virgo code bases and I would like to be able to push those changes into production.

astrofrog:
@lpsinger - yes, I can do that tomorrow, sorry for the delay! I was actually planning on having a full major release, which we are overdue for. Are you relying on the healpix part of the code? If so, is there any chance you could test out the latest developer version to see if it works properly for you? (now that we've switched from healpy to astropy-healpix).

lpsinger:
Thank you! Yes, I am specifically using the HEALPix part of the code. I am excited to try out astropy-healpix, although a simple bug fix release would be more useful as I have not yet packaged astropy-healpix for our computing clusters (Debian and RPM packages needed).

astrofrog:
@lpsinger - ok, sorry for the delay, I might do both a bugfix and a major release in that case. I'll do it tomorrow.

astrofrog:
@lpsinger - I just released v0.3.2 which just includes this fix. I'll release v0.4 later too.

lpsinger:
> @lpsinger - I just released v0.3.2 which just includes this fix.

Thank you! Just updated in MacPorts.

PR #136, opened 2017-08-30
New function for determining optimal WCS/shape for set of images

astrofrog:
This adds a new function for determining the optimal WCS/shape for a mosaic of images (or for a single image if changing coordinate frame). This is an alternative to the implementation of a a similar function in https://github.com/astrofrog/reproject/pull/125, but the version here takes into account different coordinate frames and can automatically rotate images if needed.

What is missing here compared to https://github.com/astrofrog/reproject/pull/125 is that we should use the minimal image excluding NaNs as input rather than the full images.

### DIFF
--- a/reproject/mosaicking.py
+++ b/reproject/mosaicking.py
@@ -1,14 +1,14 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
-from __future__ import (absolute_import, division, print_function,
-                        unicode_literals)
+from __future__ import absolute_import, division, print_function
 
 import numpy as np
 from astropy import units as u
 from astropy.coordinates import SkyCoord, ICRS
 from astropy.wcs.utils import (pixel_to_skycoord, skycoord_to_pixel,
-                               celestial_frame_to_wcs, proj_plane_pixel_scales)
+                               proj_plane_pixel_scales)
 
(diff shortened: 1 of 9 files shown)


astrofrog:
@keflavich @mhardcastle - would you like to review this?

reproject/mosaicking.py
```diff
@@ -0,0 +1,199 @@
...
+        x1, y1, x2, y2 = xr[order[0]], yr[order[0]], xr[order[1]], yr[order[1]]
+
+        # Determine angle between two of the vertices. It doesn't matter which
+        # ones they are, we just want to know how far from being straight the
+        # rectangle is.
+        angle = np.arctan2(y2 - y1, x2 - x1)
+
+        # Determine the smallest angle that would cause the rectangle to be
+        # lined up with the axes.
+        angle = angle % 90
```
keflavich: `np.arctan2` returns something in units of radians, not degrees, right?

keflavich: unless x1, x2, y1, and y2 are quantities?

astrofrog: Good catch!

astrofrog: They aren't quantities - this was a real bug

PR #143, opened 2017-11-13
Rebinning with and without WCS

brechmos-stsci:
This PR implements rebinning on data that includes or may not include WCS information.  The sub-sampling factor can be any floating point value.  

Two simple tests are included, one that tests the rebinning without WCS and one that tests the rebinning in the presence of the WCS.

Ran a basic test script and output is:
image

[runme_test_rebin.py.txt](https://github.com/astrofrog/reproject/files/1468096/runme_test_rebin.py.txt)

RFC @astrofrog


astrofrog:
Sorry that we never wrapped this up! Coming back at this a couple of years later I think that the right place for this to go is astropy.nddata since we already have functionality there to rebins/upsample by integer values - and this would just be an extension of that to floating point values. I'll therefore close this PR for now.

PR #144, opened 2018-01-21
Require astropy >= 1.0

lpsinger:
Astropy 1.0 introduced the function wcs_to_celestial_frame.


setup.py
```diff
@@ -106,7 +106,7 @@
       version=VERSION,
       description=DESCRIPTION,
       scripts=scripts,
-      install_requires=['numpy', 'astropy', 'astropy-healpix', 'scipy'],
+      install_requires=['numpy', 'astropy >= 1.0', 'astropy-healpix', 'scipy'],
```
astrofrog: I think the spaces around the >= should be removed

lpsinger: Done

Issue #88, opened 2015-08-18
Improve parse_output_projection


astrofrog:
Two improvements need to be made to `parse_output_projection`:
- GIve an error if the shape is an empty tuple (can happen if `NAXIS=0`)
- Have an else clause to catch other cases where e.g. filename is passed.

jimboH:
@astrofrog hi, I am interested in GSoC this year. Is this issue still open so that I could make contribution before application starts?

PR #150, opened 2018-03-07
Improve parse_output_projection

jimboH:
#88 Deal with the first improvement.
It seems like the second one has already been implemented.

### DIFF
--- a/reproject/utils.py
+++ b/reproject/utils.py
@@ -40,8 +40,8 @@ def parse_output_projection(output_projection, shape_out=None):
         except KeyError:
             if shape_out is None:
                 raise ValueError("Need to specify shape since output header does not contain complete shape information")
-        if output_projection['NAXIS'] == 0:
-            raise ValueError("The shape of the output header should not be an empty tuple")
+		if output_projection['NAXIS'] == 0:
+			raise ValueError("The shape of the output header should not be an empty tuple")
     elif isinstance(output_projection, WCS):
         wcs_out = output_projection
         if shape_out is None:
--- a/reproject/utils.py
+++ b/reproject/utils.py
@@ -1,5 +1,5 @@
 import numpy as np
-import sys
+from six import string_types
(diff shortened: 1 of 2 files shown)


reproject/utils.py
```diff
@@ -44,7 +44,15 @@ def parse_output_projection(output_projection, shape_out=None):
         wcs_out = output_projection
         if shape_out is None:
             raise ValueError("Need to specify shape when specifying output_projection as WCS object")
+    elif isinstance(output_projection, string_type):
```
astrofrog: you can replace ``string_type`` by ``six.string_types``

reproject/utils.py
```diff
@@ -44,7 +44,15 @@ def parse_output_projection(output_projection, shape_out=None):
...
+        shape_out = hdu_list[0].data.shape
+        header = hdu_list[0].header
+        wcs_out = WCS(header)
+        hdu_list.close()
     else:
-        raise TypeError('output_projection should either be a Header or a WCS object')
+        raise TypeError('output_projection should either be a Header, a WCS object, or a filename')
 
+    if len(shape_out) == 0:
+        raise ValueError("The shape of the output header should not be an empty tuple")
```
astrofrog: output header -> output image

jimboH:
@astrofrog yes I would like to add entry to CHANGES.md. Under what milestone should I add the entry? (0.5?)

reproject/utils.py
```diff
@@ -1,5 +1,5 @@
 import numpy as np
-import sys
+from six import string_types
```
astrofrog: six is already imported below:

```
 from astropy.extern import six
```

so just use ``six.string_types`` and don't add an import

jimboH: Thank you! I missed that line of code...

astrofrog:
Yes 0.5 is the correct place to add it, thanks!

PR #162, opened 2018-09-06
Add Drizzle support

### LINKED ISSUE #56: Implement support for Drizzle
cc @perrygreenfield

jamienoss:
I've been sitting on this for a couple weeks now but thought it better to throw it your way as a super initial draft.

It's dependent on the code in https://github.com/jamienoss/drizzle/pull/2

There are bits that def need cleaning up, like the ``_naxis`` refs used as a workaround for drizzle.

The tests seem to be blocked by:
 * https://github.com/astropy/astropy/issues/4973
 * https://github.com/astropy/astropy/issues/7804

Let me know if this is even what you were after/expecting?

Resolves #56


jamienoss:
FYI: @eteq & @hcferguson

astrofrog:
@jamienoss - this looks good overall, thanks! I still plan on doing a line-by-line review once the drizzle and astropy changes are in place. A few specific comments for now:

* I'd like to avoid including large FITS files for the testing. The test files are currently ~10Mb, so can you instead make them ~100kb or so? (in my experience, it's rare to find buts with a 10Mb image that you wouldn't see with a 100kb image)

* I'd like to avoid having to require the latest developer version of astropy, so can you (for now) include a workaround for the zero-sized arrays, so that we can remain compatible with older astropy versions? (and we can then remove the workaround later)

PR #166, opened 2018-12-05
Implement support for APE14

astrofrog:
This implements full support for reprojection with [APE14](https://github.com/astropy/astropy-APEs/blob/master/APE14.rst)-compliant WCS objects. A few points to consider:

* The utilities in ``wcs_utils.py`` are actually pretty general and could potentially be included in ``astropy.wcs.wcsapi``. The main function here is one that given pixel coordinates and two WSCes will do a pixel -> world -> pixel transformation, but being memory-efficient in cases where there are uncorrelated axes. What do people think -  should this remain in reproject, or should we try and include in the core package?

* Currently I've bumped the minimum required version to Astropy 3.2, which means dropping Python 2.7 support. Are we ok with that, and is Astropy 3.2 too strict a requirement?

Once this is merged, I think I'll email astropy-dev to ask for people to take it for a test spin, since it's a reasonably big change and I want to make sure we haven't broken anything.

Fixes https://github.com/astropy/reproject/issues/182


astrofrog:
This PR accidentally got rid of the round-tripping check which is important for e.g. solar data, so I'll need to add a regression test before merging this to make sure I don't break this again in future.

keflavich:
@astrofrog was it intentional to delete the functionality iterating over celestial slices?  That functionality appears to be completely lost now

Note that this change broke spectral-cube.

astrofrog:
@kevlavich - for the interpolating reprojection it will do that by default now (the logic for that has been generalized to any non-correlated axes). Can you point me to the broken spectral-cube functionality?

keflavich:
In spectral-cube, we were passing along the `independent_celestial_slices` parameter, whcih is no longer accepted.  Removing that parameter is easy, but I think we need to then have that be reproject-version-dependent, since the default was `independent_celestial_slices=False`

Does the current version really replace that logic?  One of the hidden advantages of the iterate-over-slices approach was that it limited the instantaneous memory footprint.  Directly using `map_coordinates` will necessarily load the whole cube at once, no?

astrofrog:
Ah I should probably deprecate the ``independent_celestial_slices`` argument though so that packages can be compatible with this and older versions of reproject.

PR #170, opened 2019-01-11
Fix malformatted .. doctest-skip:: directives in Very Large Cubes docs

keflavich:
see https://reproject.readthedocs.io/en/latest/celestial.html#very-large-cubes.  The problem was having `::` before the `.. doctest-skip::` directive.


astrofrog:
Thanks!

Issue #180, opened 2019-06-18
reproject.mosaicking missing from API docs


astrofrog:
It looks like we are missing reproject.mosaicking from the API documentation - this will need to be included at the bottom of ``index.rst`` in the docs.

PR #186, opened 2019-07-28
Added new reproject_and_coadd function

astrofrog:
While one a ✈️ today I finally took a stab at adding a ``reproject_and_coadd`` function which is the missing piece for doing mosaicking. This includes an algorithm to match backgrounds between images. Here's a preview of a 2MASS mosaic:

mosaicking-2

and a comparison of the differences without and with background matching (there are other datasets for which this would be more obvious but it's still somewhat visible here):

mosaicking-4

Remaining TODOs:

* [x] Add more tests, in particular for the background matching solver
* [x] Add changelog entry
* [x] Added note about accuracy of photometry etc.
* [x] Update docs to use data from a remote source

Things to do either as part of this PR or after:

* [ ] Use the footprint from each image as a weight in the averaging (will matter for reproject_exact)
* [ ] Allow footprints as inputs to use as weight maps (could be useful for e.g. radio reprojection)
* [ ] Support n-dimensional data

Fixes https://github.com/astropy/reproject/issues/180


astrofrog:
@keflavich @bsipocz @cdeil @lpsinger @larrybradley @mhardcastle @Cadair - as people who've either committed to this repo in the past or related functionality in the astropy ecosystem, it would be great if some of you would be willing to take a look at this to see if you like the API (and optionally do a more in-depth code review). If you have time for a cursory review, then it probably makes sense to mostly look at the new ``mosaicking.rst`` page. Thanks! 🙏

docs/mosaicking.rst
```diff
@@ -0,0 +1,310 @@
...
+    EQUINOX =               2000.0 / [yr] Equinox of equatorial coordinates
+    >>> shape_out
+    (2201, 1800)
+
+Coordinate system
+-----------------
+
+By default, the coordinate system of the first file is used, and the final
+WCS is set up so that North (in that coordinate system) is up. In the
+case above, the images are in equatorial coordinats, so the final WCS is also
```
keflavich:
```suggestion
case above, the images are in equatorial coordinates, so the final WCS is also
```

docs/mosaicking.rst
```diff
@@ -0,0 +1,310 @@
...
+in equatorial coordinates. We can force the output WCS to instead be in
+Galactic coordinates by setting the ``frame=`` argument to a coordinate frame
+object such as :class:`astropy.coordinates.Galactic`::
+
+    >>> from astropy.coordinates import Galactic
+    >>> wcs_out, shape_out = find_optimal_celestial_wcs(m17_hdus,
+    ...                                                 frame=Galactic())
```
keflavich: do strings work?  e.g., `'fk5'`, `'icrs'`, `'galactic'`?  For consistence with the `astropy.coordinates` API, they should

docs/mosaicking.rst
```diff
@@ -0,0 +1,310 @@
...
+    >>> shape_out
+    (1800, 2201)
+
+As expected, the optimal shape is smaller than was returned previously.
+
+Resolution
+----------
+
+By default, the final mosaic will have the resolution of the highest resolution
+input image, but this can be overriden using the ``resolution=`` keyword
```
keflavich: maybe describe what `resolution` means more specifically.  I assume it is the pixel scale along the pixel axes?  (i.e., not diagonal to the pixel, or anything weird like that)

Actually, if it _is_ pixel scale, I prefer using that term over `resolution`, since we typically use `resolution` to refer to the telescope-optics-produced PSF scale rather than the pixel scale.

docs/mosaicking.rst
```diff
@@ -0,0 +1,310 @@
...
+    >>> wcs_out, shape_out = find_optimal_celestial_wcs(m17_hdus,
+    ...                                                 resolution=1.5 * u.arcsec)
+
+Projection and reference coordinate
+-----------------------------------
+
+Finally, you can customize the projection to use as well as the reference
+coordinate. To change the projection from the default (which is the
+gnomonic projection, or ``TAN``), you can use the ``projection=`` keyword
+argument, which should be set to a valid three-letter FITS-WCS projection code::
```
keflavich: link this to http://adsabs.harvard.edu/abs/2002A%26A...395.1061G?

docs/mosaicking.rst
```diff
@@ -0,0 +1,310 @@
...
+:func:`~reproject.reproject_interp` should also be specified, and be
+given the shape of the output image using the Numpy ``(ny, nx)`` convention
+(this is because :class:`~astropy.wcs.WCS`, unlike
+:class:`~astropy.io.fits.Header`, does not contain information about image
+size).
+
+Finally, the ``reproject_function`` should be used to specify which function to
+use to reproject individual tiles - this should be either
+:func:`~reproject.reproject_interp` or :func:`~reproject.reproject_exact` - with
+the latter being slower but more accurate. Keyword arguments for these functions
```
keflavich: maybe emphasize `"...accurate and necessary for flux conservation."` ?

docs/mosaicking.rst
```diff
@@ -0,0 +1,310 @@
...
+    im1 = ax1.imshow(array, origin='lower', vmin=600, vmax=800)
+    ax1.set_title('Mosaic')
+    ax2 = plt.subplot(1, 2, 2)
+    im2 = ax2.imshow(footprint, origin='lower')
+    ax2.set_title('Footprint')
+
+In some cases, including the above example, each tile that was used to compute
+the mosaic has an arbitrary offset due e.g. to different observing conditions.
+The :func:`~reproject.mosaicking.reproject_and_coadd` includes an option to
+match the backgrounds (assuming a constant additive offset in each image)::
```
keflavich: comment on the algorithm used for background matching and whether it is adjustable

docs/mosaicking.rst
astrofrog: Fixed

docs/mosaicking.rst
astrofrog: Ok I've tried to clarify things here - if we want to change the name of the actual keyword argument, we should do that in a separate PR with a deprecation phase, because that function (find_optimal...) is already in a released version of reproject (I'm just adding the docs for that here)

docs/mosaicking.rst
astrofrog: Done

docs/mosaicking.rst
astrofrog: Done

docs/mosaicking.rst
astrofrog: I've added a paragraph and mentioned that there are no exposed arguments for controlling this for now. The internal SGD function I wrote does have some options but we don't expose this at a higher level, and I'm not convinced we need to, but we can always re-assess in future once this is actually used in production?

astrofrog:
@keflavich - thanks for the docs review! I've pushed some updates.

pllim:
cc @mcara , @stsci-hack , and @jdavies-st -- please pass this along to interested parties at STScI.

@astrofrog , if you want to test background matching more, try some WFPC2 data from HST.

reproject/mosaicking/background.py
```diff
@@ -0,0 +1,133 @@
...
+from math import exp
+
+import numpy as np
+
+__all__ = ['solve_corrections_sgd', 'determine_offset_matrix']
+
+
+def determine_offset_matrix(arrays):
+    """
+    Given a list of ReprojectedArraySubset, determine the offset
```
mcara: What are `ReprojectedArraySubset`? I can't find their definition.

astrofrog: See the ``subset_array.py`` file - this is not meant to be part of the public API.

fardal:
Does it work on images with NaNs, indicating areas to ignore?  I tried on some of these and it gave an error.

astrofrog:
@fardal - I just tried with some data containing NaN and it seemed to work. Would you be happy to share the example you were trying so I can investigate?

astrofrog:
@fardal - I found some issues with images with NaN values and have now fixed those.

astrofrog:
I believe this works reasonably well and I would like to try and get more people to try this out so I'm going to go ahead and merge, and will then advertise this functionality on astropy-dev for people to try. I'll continue improving this in subsequent pull requests.

PR #189, opened 2019-08-28
Added python_requires to setup.py

### LINKED ISSUE #178: setup.py should define python_requires
Currently we don't specify which versions of Python we are compatible with - this should be done with ``python_requires`` in ``setup.py`` (or in ``setup.cfg`` if using that)

astrofrog:
Fixes #178


astrofrog:
The failures are unrelated.

PR #205, opened 2019-11-08
Split out interpolation WCS validation into new function

dstansby:
This is purely a code aesthetic change, but I think it makes it much more readable with the validation split off into it's own function.


astrofrog:
@dstansby - thanks, I agree this makes sense. There are some PEP8 failures though, could you fix these? (``tox -e pep8`` to reproduce locally)

astrofrog:
Thanks @dstansby!

PR #208, opened 2019-11-21
Fix reproject_and_coadd for cyclic coordinate systems

dstansby:
`reproject_and_coadd` is clever and tries to minimize the footprint that is reprojected. However, the (0, 0) pixels it was using were actually the centers of the (-1, -1) pixels (-0.5, -0.5), which for cyclic coordinate systems (e.g. Heliographic Carrington) would wrap round to the top right corner, and therefore the footprint to be re-projected was completely empty!

This (I think) fixes that by using the edges of the corner pixels instead of the pixel centers.

I am struggling to produce a simple and self contained example...

### DIFF
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -117,8 +117,8 @@ def reproject_and_coadd(input_data, output_projection, shape_out=None,
         # distortions of the edges in the reprojection process we could simply
         # add arbitrary numbers of midpoints to this list.
         ny, nx = array_in.shape
-        xc = np.array([-0.5, nx - 0.5, nx - 0.5, -0.5])
-        yc = np.array([-0.5, -0.5, ny - 0.5, ny - 0.5])
+        xc = np.array([0, nx + 1, nx + 1, 0])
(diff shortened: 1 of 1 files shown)


astrofrog:
Just to check I understand - integer positions are at the center of pixels, so (-0.5, -0.5) is the edge of the bottom left pixels, so in this case it looks like you are changing to the center of the pixels?

astrofrog:
I do remember stable sunpy having an 0.5 offset bug in some of the header machinery though, which was fixed recently - @Cadair, can you clarify?

dstansby:
> Just to check I understand - integer positions are at the center of pixels, so (-0.5, -0.5) is the edge of the bottom left pixels, so in this case it looks like you are changing to the center of the pixels?

Ah I must have misunderstood and jumped the gun a bit here. It seems like (-0.5, -0.5) is what is wanted, to get the edge of the pixels, but I guess the problem is that that pixel coordinate roundtrips from the bottom LH corner (-0.5, -0.5) to the bottom RH corner (nx + 0.5, -0.5) in a Heliographic Carrington frame, I guess because of the way the coordinates are wrapped.

I will try and put together an example (and therefore test).

PR #214, opened 2020-01-16
Add support for blocked and parallel reprojection in ``reproject_interp``

AlistairSymonds:
Performing reprojection of any non-healpix type by breaking up the out space into blocks and iterating over each block to perform reprojection into that specific block of area. This means a large output space can be used whilst only having to load one block into memory at a time. 

Additionally these output blocks can be processed in parallel to achieve a speed up of any function. 

As mentioned there is still some debug code in there which needs to be cleaned up (commented print statement, imports used for profiling), but the functionality is there.

### DIFF
--- a/reproject/interpolation/high_level.py
+++ b/reproject/interpolation/high_level.py
@@ -1,8 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
-
+import os
 from astropy.utils import deprecated_renamed_argument
 
-from ..utils import parse_input_data, parse_output_projection
+from ..utils import parse_input_data, parse_output_projection, reproject_blocked
 from .core import _reproject_full
 
 __all__ = ['reproject_interp']
@@ -17,15 +17,11 @@ ORDER['bicubic'] = 3
 @deprecated_renamed_argument('independent_celestial_slices', None, since='0.6')
 def reproject_interp(input_data, output_projection, shape_out=None, hdu_in=0,
                      order='bilinear', independent_celestial_slices=False,
-                     output_array=None, return_footprint=True,
-                     roundtrip_coords=True):
+                     output_array=None, return_footprint=True, output_footprint=None,
+                     block_size=None, parallel=False, roundtrip_coords=True):
     """
     Reproject data to a new projection using interpolation (this is typically
     the fastest way to reproject an image).
-
-    The output pixel grid is transformed to the input pixel grid, and the
-    data values in ``input_data`` interpolated on to these coordinates to get
-    the reprojected data on the output grid.
 
     Parameters
     ----------
@@ -69,6 +65,11 @@ def reproject_interp(input_data, output_projection, shape_out=None, hdu_in=0,
         extremely large files.
     return_footprint : bool
         Whether to return the footprint in addition to the output array.
+    block_size : None or tuple of (int, int)
+        If not none, a blocked projection will be performed where the output space is
+        reprojected to one block at a time, this is useful for memory limited scenarios
+        such as dealing with very large arrays or high resolution output spaces.
+    parallel : bool or int
     roundtrip_coords : bool
         Whether to verify that coordinate transformations are defined in both
         directions.
@@ -90,6 +91,26 @@ def reproject_interp(input_data, output_projection, shape_out=None, hdu_in=0,
     if isinstance(order, str):
         order = ORDER[order]
 
-    return _reproject_full(array_in, wcs_in, wcs_out, shape_out=shape_out,
-                           order=order, array_out=output_array,
-                           return_footprint=return_footprint, roundtrip_coords=roundtrip_coords)
+    # if either of these are not default, it means a blocked method must be used
+    if block_size is not None or parallel is not False:
+        # if parallel is set but block size isn't, we'll choose
+        # block size so each thread gets one block each
+        if parallel is not False and block_size is None:
+            block_size = shape_out.copy()
+            # each thread gets an equal sized strip of output area to process
+            block_size[0] = shape_out[0] // os.cpu_count()
+
+        # given we have cases where modern system have many cpu cores some sanity clamping is
+        # to avoid 0 length block sizes when num_cpu_cores is greater than the side of the image
+        for dim_idx in range(min(len(shape_out), 2)):
+            if block_size[dim_idx] == 0:
+                block_size[dim_idx] = shape_out[dim_idx]
+
+        return reproject_blocked(_reproject_full, array_in=array_in, wcs_in=wcs_in, wcs_out=wcs_out,
+                                 shape_out=shape_out, output_array=output_array, parallel=parallel,
+                                 block_size=block_size, return_footprint=return_footprint,
+                                 output_footprint=output_footprint)
+    else:
+        return _reproject_full(array_in, wcs_in, wcs_out, shape_out=shape_out,
+                               order=order, array_out=output_array,
+                               return_footprint=return_footprint, roundtrip_coords=roundtrip_coords)
--- a/reproject/interpolation/high_level.py
+++ b/reproject/interpolation/high_level.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import os
+
 from astropy.utils import deprecated_renamed_argument
 
 from ..utils import parse_input_data, parse_output_projection, reproject_blocked
@@ -70,6 +71,11 @@ def reproject_interp(input_data, output_projection, shape_out=None, hdu_in=0,
         reprojected to one block at a time, this is useful for memory limited scenarios
         such as dealing with very large arrays or high resolution output spaces.
     parallel : bool or int
+        Flag for parallel implementation. If ``True``, a parallel implementation
+        is chosen, the number of processes selected automatically to be equal to
+        the number of logical CPUs detected on the machine. If ``False``, a
+        serial implementation is chosen. If the flag is a positive integer ``n``
+        greater than one, a parallel implementation using ``n`` processes is chosen.
     roundtrip_coords : bool
         Whether to verify that coordinate transformations are defined in both
         directions.
--- a/reproject/interpolation/tests/test_core.py
+++ b/reproject/interpolation/tests/test_core.py
@@ -586,13 +586,24 @@ def test_identity_with_offset(roundtrip_coords):
 @pytest.mark.parametrize('parallel', [True, 0, 2, False])
 @pytest.mark.parametrize('block_size', [[10, 10], [500, 500], [500, 100], None])
 def test_blocked_against_single(parallel, block_size):
+
+    # Ensure when we break a reprojection down into multiple discrete blocks
+    # it has the same result as if all pixels where reprejcted at once
+
+    hdu1 = fits.open(get_pkg_data_filename('galactic_center/gc_2mass_k.fits'))[0]
+    hdu2 = fits.open(get_pkg_data_filename('galactic_center/gc_msx_e.fits'))[0]
+    array_test = None
+    footprint_test = None
+
     # the warning import and ignore is needed to keep pytest happy when running with
     # older versions of astropy which don't have this fix:
     # https://github.com/astropy/astropy/pull/12844
     # All the warning code should be removed when old version no longer being used
+    # Using context manager ensure only blocked function has them ignored
     import warnings
-    warnings.simplefilter('ignore', category=FITSFixedWarning)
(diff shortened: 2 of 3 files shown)


astrofrog:
Thanks for working on this! I've been taking a look at it and thinking about what would make sense from the point of view of the user API - in particular I think that we could simplify things a little if we don't expect this function to be called by the user directly, but having the user set e.g. ``chunk_size=(100,100)`` (and optionally ``parallel=...``) when calling ``reproject_interp`` and ``reproject_exact``, and then in the ``high_level.py`` files, once the input has all been parsed and sanitized, e.g here:

https://github.com/astropy/reproject/blob/0ed1b40b881a98440f9d79d958a29c884c3a5ad1/reproject/interpolation/high_level.py#L84

we would check if ``chunk_size`` is set, and if so, call your helper function to call the core function over chunks. This would mean not needing to do any parsing/validation in your function and also a simple API for users. Would you have time to explore this? If not, I'm happy to take over the PR and add commits to it, so just let me know what is best for you :)

AlistairSymonds:
I'm happy to look at it, just making sure I understand correctly, you mean doing something like this:

https://github.com/AlistairSymonds/reproject/blob/b7ab13c800b71526e3b38af8e118fcdf3edd6798/reproject/interpolation/high_level.py#L95-L101

astrofrog:
Yes exactly!

AlistairSymonds:
Okay that let me clean up the blocked_reproject function but there are still some issues where the interfaces for core functions don't quote match up, for instance both adaptive and interp have the `order` argument but the exact function does not.

AlistairSymonds:
@astrofrog any thoughts? It's cleaned up the issue a bit but not entirely (if the functions took **kwargs maybe it could be done nicer?)

astrofrog:
@AlistairSymonds - I'll look at this shortly, I'm very sorry for the delay!

AlistairSymonds:
Hey @astrofrog I've found some free time up my sleeve and a need to reproject a large number of images. So I've gone through and cleaned up some of the issues left dangling so all the defaults tests are passing.

There is two things remaining I think though, 

- obviously some new tests are needed for the parallel functionality. I was just going to take one of the examples from say the tutorial and compare the output of blocked/parallel against existing single thread whole image impl. Any comments you want to add on that plan would be appreciated
 
- The round trip tests are skipped unless sunpy is installed, the CI doesn't run these tests, are they expected to be ran or deprecated/purposefully not a priority?


I've got a few other ideas in the back of my mind that should really help speed up reprojecting large areas of the sky at high resolutions like culling empty tiles, but its the kind of thing that can probably wait until after this PR :)

AlistairSymonds:
> The round trip tests are skipped unless sunpy is installed, the CI doesn't run these tests, are they expected to be ran or deprecated/purposefully not a priority?

Apologies seems I was misunderstanding the AzurePipe reports, and they indeed ran and passed when I properly update my local tree to merge in latest from this repo.

I also only just found the astropy top level contributing guide, its highly likely it's got plenty of suggested changes too. (might be worth having a contributing file in this repo that just points there for dummies like me who aren't perfectly across astropy?)

pllim:
The CI is a little busted right now. It is unfortunate that this PR has been open for so long. We apologize for any inconvenience caused. Hopefully @astrofrog will come back to reviewing this soon...

reproject/interpolation/high_level.py
```diff
@@ -82,5 +92,23 @@ def reproject_interp(input_data, output_projection, shape_out=None, hdu_in=0,
...
         order = ORDER[order]
 
-    return _reproject_full(array_in, wcs_in, wcs_out, shape_out=shape_out, order=order,
-                           array_out=output_array, return_footprint=return_footprint)
+    # if either of these are not default, it means a blocked method must be used
+    if block_size is not None or parallel is not False:
+        # if parallel is set but block size isn't, we'll choose
+        # block size so each thread gets one block each
+        if parallel is not False and block_size is None:
+            block_size = [dim // os.cpu_count() for dim in shape_out]
```
keflavich, line 100: This code block doesn't match the comment above, right?  If we have 4 CPUs, say, and a 2D image, `block_size` will be 1/16th the image size, so each CPU will get 4 blocks.  This isn't necessarily a bad thing, but it's also not obviously the right default.

reproject/utils.py
```diff
@@ -133,3 +136,103 @@ def parse_output_projection(output_projection, shape_out=None, output_array=None
...
+    if output_array is None:
+        output_array = np.zeros(shape_out, dtype=float)
+    if output_footprint is None and return_footprint:
+        output_footprint = np.zeros(shape_out, dtype=float)
+
+    # setup variables needed for multiprocessing if required
+    proc_pool = None
+    blocks_futures = []
+
+    if parallel or type(parallel) is int:
```
keflavich, line 170: I think we want `and` here?  Otherwise `parallel=0` will try to work with 0 workers

keflavich (approved):
I have two really minor comments but this looks great to me.  I'm testing it out on some cubes now.

keflavich:
One more aside - it would be nice to add some sort of progressbar support, but that can be a different Issue.

reproject/interpolation/high_level.py
AlistairSymonds: Ah yes you're dead right, and its probably worth doing big long strips that match the fits memory layout anyway.

reproject/utils.py
AlistairSymonds: This might need tightening up, I was imagining parellel=True as being let reproject auto determine threading, but as you say zero would lead to issues.

AlistairSymonds:
Also just quickly while there's more people looking at this, I'm gonna go over @keflavich's comments and make some updates if no one says otherwise. 

Generally if anyone has feedback on more higher level python/astropy ways of doing things, I'll take it, I'm someone who's written a fair bit of python at various points but main wheel house is verilog and HW design - so I've probably missed something obvious.

astrofrog (changes requested):
@AlistairSymonds - this is looking good so far! Please do go ahead and rebase (to hopefully fix the CI) and implement @keflavich's comments.

The only comment I have for now is a general one, which is that I think it would be good to change the terminology from 'block' to 'chunk' to match dask terminology - as I think we might want to add support for dask anyway in reproject in future (for input/output arrays) and it would be confusing to have two different terms.

I have more time available to work now so I will be more responsive here now and would like to make sure we can include this in the next release! 😄

AlistairSymonds:
Yeah I've got no strong attachment to any term, there's lots of options that are as equally overloaded as each other it seems - tiles, chunks, block, bins.

And frick I've completely beans'd up the rebase, I still had a master branch laying about in my fork instead of main and its all gone awry - gonna fix that up then address the comments for realsies this time instead of just talking about it.

AlistairSymonds:
Okay as nice as it would be to have the original 3yr old pull request be accepted the git history is turning into a nightmare - when I have time on the weekend I'm going to make a new devbranch then manually port across. (Is it obvious I've only used perforce professionally? :P )

AlistairSymonds:
@astrofrog I'd say its 99% clean, the only issue left seems to be from the testdata being used when I run with the oldestdeps config. It seems something to do with (un)pickling the WCS causing a FITSFixedWarning in unpickling when returning from the multiprocessing call of _block() causing pytest to fall apart?


The only thing I've found otherwise is the following two pull requests

https://github.com/astropy/astropy/pull/12844

which came from it being identified as a bug here: https://github.com/astropy/astropy/issues/12834

So indeed the lines causing the warning have been fixed in later astropy versions, and when running the test code outside of we just get warnings but the np assert doesn't fire, eg:



Is this something worth working around or just leave as is since a newer version of astropy has fixed it?

pllim:
Since it has been fixed in astropy already, you could use `warnings` module to ignore this warning for the test here. Thanks!

reproject/interpolation/tests/test_core.py
```diff
@@ -586,6 +586,25 @@ def test_identity_with_offset(roundtrip_coords):
 @pytest.mark.parametrize('parallel', [True, 0, 2, False])
 @pytest.mark.parametrize('block_size', [[10, 10], [500, 500], [500, 100], None])
 def test_blocked_against_single(parallel, block_size):
+    # the warning import and ignore is needed to keep pytest happy when running with
+    # older versions of astropy which don't have this fix:
+    # https://github.com/astropy/astropy/pull/12844
+    # All the warning code should be removed when old version no longer being used
+    import warnings
+    warnings.simplefilter('ignore', category=FITSFixedWarning)
```
pllim, line 594: Does this ignore it for the whole test session? Would a context manager be safer?

AlistairSymonds: I did a quick test by just adding the following function to the bottom of the test file and it still failed correctly - so no doesn't seem to affect the whole pytest session. 
```
def test_fitswarning():
    raise FITSFixedWarning
```

I assume each test gets its own new invocation of the python interepter so it works but couldn't find any conrete reference to that implementation detail or if it should be relied upon.

So I do agree it would be better if a ` with warnings.catch_warnings()` manager was used, since then it could be used just wrap the new changes - I've done this locally and just running some checks

AlistairSymonds:
I think that's everything now? Once this is in I'm also happy to take a look at doing a similar thing for the adaptive and exact methods too - just tackling one problem at a time :D

reproject/interpolation/high_level.py
```diff
@@ -69,6 +65,11 @@ def reproject_interp(input_data, output_projection, shape_out=None, hdu_in=0,
         extremely large files.
     return_footprint : bool
         Whether to return the footprint in addition to the output array.
+    block_size : None or tuple of (int, int)
+        If not none, a blocked projection will be performed where the output space is
+        reprojected to one block at a time, this is useful for memory limited scenarios
+        such as dealing with very large arrays or high resolution output spaces.
+    parallel : bool or int
```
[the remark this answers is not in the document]
AlistairSymonds: I'll go over this another time/recomment etc - I was trying to emulate the parallel argument already used in the reproject_exact() function but it's definitely been a while since I originally wrote and looked closely at this.

reproject/interpolation/high_level.py
```diff
@@ -90,6 +91,26 @@ def reproject_interp(input_data, output_projection, shape_out=None, hdu_in=0,
...
-                           order=order, array_out=output_array,
-                           return_footprint=return_footprint, roundtrip_coords=roundtrip_coords)
+    # if either of these are not default, it means a blocked method must be used
+    if block_size is not None or parallel is not False:
+        # if parallel is set but block size isn't, we'll choose
+        # block size so each thread gets one block each
+        if parallel is not False and block_size is None:
+            block_size = shape_out.copy()
+            # each thread gets an equal sized strip of output area to process
+            block_size[0] = shape_out[0] // os.cpu_count()
```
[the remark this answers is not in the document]
AlistairSymonds: Block size is controllable separately to allow for memory usage tuning, eg if you're reprojecting a field 10s of degrees wide to a 0.5"/px scale that just blows up memory usage. This was my original usecase for doing this blocked functionality - everything possible to stop loading the entire input and output arrays into memory at once. 

Additonally there's also some existing code in the moasicking reproject_and_coadd() function that does some culling on a full input field granularity - I'm not sure if it's actually robust with all distortions/projections but ideally a similar sort of empty block culling could be used in future here too. 

Happy to change the parallel=1 case too.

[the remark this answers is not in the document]
astrofrog: Part of me thinks parallel=1 should actually use the parallel infrastructure rather with one worker rather than be the same as False - as it could be useful for debugging?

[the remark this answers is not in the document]
AlistairSymonds: That was my original thought too - or when someone wants the maths to happen in another process so they could do GIL shennigans in the main thread or something? Not super attached either way and not a big change to do, I've got tomorrow off of work so can address it along with the other stuff Lim mentioned if needed.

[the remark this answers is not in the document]
astrofrog: Ok - I would suggest leaving the current behavior as-is.

astrofrog:
@AlistairSymonds - if you can update the docstrings as @pllim mentioned and avoid the use of the continuation character, I'll go ahead and merge this. Thanks!

astrofrog (approved):

astrofrog:
@AlistairSymonds - thanks for your work on this and sorry it took so long to merge!

AlistairSymonds:
Cheers! Glad it happened in the end - I'll take a look at doing the same for adaptive and exact methods now a base has been put in.

PR #234, opened 2020-05-28
Fix compatibility with astropy 4.1rc1

astrofrog:
There are several fixes here:

* I am removing an old unused file that was causing CI to fail due to the recent changes in pytest-doctestplus that cause all Python files to now be imported (this tripped up on an old wcsaxes import)
* The mosaicking.rst page needed to be updated due to the changes in io.votable related to string columns (https://github.com/astropy/astropy/pull/9505) - the code now is compatible with astropy pre and post 4.1. Once only 4.1+ is supported, the code can be simplified a little.

There is one more failure related to Sunpy and SkyCoord equality checks. This appears to be fixed in SunPy 2.0rc2 so no further action required there.


astrofrog:
The failures are unrelated and will need to be investigated separately.

Issue #230, opened 2020-05-05
Weights images for reproject_and _coadd()


adebahr:
I try to combine radio images to a mosaic using the input images plus their weights. Both types of images are in the standard fits format, are on the same grid and contain the same header information.
I first calculate the optimal wcs using:

all_2Dfiles = sorted(glob.glob(imagedir + '/B[0-9][0-9]_pb.fits'))
all_hdus = [pyfits.open(image)[0] for image in all_2Dfiles]
wcs_opt, shape_opt = find_optimal_celestial_wcs(all_hdus)

And then I generate the image giving the weights as a second list of fits-files:

input_weights = sorted(glob.glob(imagedir + '/B[0-9][0-9]_wg.fits'))
array, footprint = reproject_and_coadd(all_hdus, wcs_opt, input_weights=input_weights, shape_out=shape_opt, reproject_function=reproject_interp)

I then get the following error message:

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-19-ad3eed9d33f7> in <module>
      2 
      3 input_weights = sorted(glob.glob(imagedir + '/B[0-9][0-9]_wg.fits'))
----> 4 array, footprint = reproject_and_coadd(all_hdus, wcs_opt, input_weights=input_weights, shape_out=shape_opt, reproject_function=reproject_interp)

/home/apercal/pipeline/lib/python3.7/site-packages/reproject/mosaicking/coadd.py in reproject_and_coadd(input_data, output_projection, shape_out, input_weights, hdu_in, reproject_function, hdu_weights, combine_function, match_background, background_reference, **kwargs)
    107             weights_in = None
    108         else:
--> 109             weights_in = parse_input_weights(input_weights[idata], hdu_weights=hdu_weights)
    110             if np.any(np.isnan(weights_in)):
    111                 weights_in = np.nan_to_num(weights_in)

/home/apercal/pipeline/lib/python3.7/site-packages/reproject/utils.py in parse_input_weights(input_weights, hdu_weights)
     45 
     46     if isinstance(input_weights, str):
---> 47         return parse_input_data(fits.open(input_weights), hdu_weights=hdu_weights)
     48     elif isinstance(input_weights, HDUList):
     49         if hdu_weights is None:

TypeError: parse_input_data() got an unexpected keyword argument 'hdu_weights'

[...]

PR #249, opened 2021-03-16
Fix implementation of hdu_weights

astrofrog:
Fixes https://github.com/astropy/reproject/issues/230

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,11 @@
   into memory all at once, since the actual data aren't needed for optimal WCS
   determination. [#242]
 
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -34,6 +34,7 @@ testall =
     asdf
     gwcs
     pyvo
+    matplotlib
```
Cadair, line 37: Is this because of the sunpy tests? If so I fixed it in #250 by adding the `[map]` extra.

astrofrog: Yes I'll remove this, thanks!

PR #257, opened 2021-11-06
Add some more docs for reproject_interp

dstansby:
I wrote these as I was reading the code trying to understand how `reproject_interp` works, and thought they were worth adding.


pllim (approved):
LGTM. Thanks!

PR #263, opened 2021-12-31
MNT: Stop using distutils

pllim:
`distutils` is deprecated.

This package does not seem to be affected by astropy/astropy#12633.


pllim:
Fixing the wheel job is out of scope here.

PR #274, opened 2022-01-19
Fix Jacobian calculation in the adaptive algorithm

svank:
I've been testing and improving the Adaptive Resampling implementation for my own work (and in collaboration with Craig, who originally published the algorithm in 2004). This is the first of three pull requests I'll be sending, to try to break up the changes and ease review. This PR is one bug fix (required by the next PR), the next PR will have significant optimizations, and the third will have a few more bug fixes and added features that achieve robust, flux-conserving resampling. (Do let me know if I should break things up differently!)

The summary for this fix is that the algorithm uses the Jacobian matrix as a local linearization of the input<->output coordinate transformation, but this implementation was actually computing the transpose of the Jacobian matrix and using it as if it were un-transposed. This means two of the four elements of this 2x2 matrix were incorrect, and this impacted the averaging weights computed for each sample taken from the input image. For this to affect the output image significantly, the Jacobian must have large off-diagonal elements, and this may mean the real-world impact of this bug was minor---it certainly is minor for the test cases implemented for this algorithm.

[Here is a notebook](https://gist.github.com/svank/1e77c383b2bcc5431a1c129b6099b475) I put together describing this in more detail, showing a manufactured example in which the effect of this bug is very clear, and showing that the impact is minimal for the test cases. Fair warning, at the top it includes my work convincing myself that this was in fact a bug and not me just mis-understanding some nuance, and that may be more in-depth than is needed for this PR.

This fix requires updating the reference images for two of the three test cases. I also added a fourth test case similar to the example in the notebook, which ensures that sort of case is being handled correctly.


Cadair:
hey @svank Thanks a lot for these PRs. Both @astrofrog and I are pretty limited on time at the moment (I am running around trying to get some stuff done before DKIST starts taking data), these PRs look amazing and I will try and find time to look at them asap :)

astrofrog (approved):
To the best of my understanding, this seems correct/reasonable. Thanks for all your work on this @svank!

PR #277, opened 2022-01-31
TST: Pin coverage for now

pllim:
Timing out because of nedbat/coveragepy#1310 .


pllim:
Failing is better than timing out. Not sure what the SHA error is about.

Azure failure is unrelated.

PR #279, opened 2022-02-03
Add boundary handling options for adaptive resampling

svank:
I lied, here's a fourth PR, applying on top of #276. The two commits belong to this PR are "Add boundary handling..." and "Set default boundary mode...".

This adds a few different boundary-handling modes, to adjust how the adaptive resampling algorithm handles the case when the sampling window straddles or lies outside of the boundary of the input image---that is, when some of the input pixels that should be sampled for a given output pixel do not exist. The existing behavior was to simply ignore those out-of-bounds samples and base the output pixel value on the remaining in-bounds samples (or set the output to NaN if there are no in-bounds samples), and that can give rise to some strong edge effects, depending on the input image. The old behavior can be had with the "ignore" mode in this PR. ~~I've set the default to be the "strict" mode, which sets output pixels to NaN if any of the input pixels to be sampled do not exist. That's a change in default behavior, but it seems like the most conservative choice and so the best default.~~ (EDIT: this defaults change has been moved to #293.) There are also "nearest", "constant", "grid-constant", and "ignore_threshold" modes, for which I'll refer to the added documentation.

The existing `deforest.pyx` had some support for treating the x or y axes of the input image as cyclic, which I'd been sort of ignoring but trying not to break in the previous PRs. I've tested that functionality and exposed it in the public API as well. There are separate flags for the two axes, and the boundary mode then applies to any non-cyclic axes.

One could imagine someday adding a similar choice of boundary modes to the interpolating and exact algorithms, though I haven't attempted to do so. The interpolating algorithm wraps `scipy.ndimage.map_coordinates` which does offer a menu of boundary modes, and so in case those are ever exposed for the interpolating algorithm, I've matched the names of my boundary modes to its menu when it implements a similar mode. (Though it seems to have cyclic axes as an all-or-nothing choice, whereas I've maintained the per-axis choice via separate flags.)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,18 @@
 
 - Drop support for Python 3.7.
 - Infrastructure and packaging updates.
+- Made many improvements, bug fixes, and significant speed-ups for the adaptive
+  resampling algorithm, ``reproject_adaptive``. These bug fixes may cause
+  changes to the reprojected images, which are typically negligible.
+  Improvements include the addition of a flux-conserving mode, support for a
+  Gaussian filter kernel, a menu of boundary-handling modes, and a
+  ``center_jacobian`` flag to trade speed for accuracy with rapidly-varying
+  transformations.
+- Added a ``roundtrip_coords`` argument to ``reproject_adaptive`` and
+  ``reproject_interp``. By default, all coordinate transformations are run in
+  both directions to handle some situations where they are ambiguous. This can
+  be disabled by setting ``roundtrip_coords=False`` which may offer a
+  significant speed increase.
 
 0.8 (2021-08-11)
 ----------------
--- a/docs/celestial.rst
+++ b/docs/celestial.rst
@@ -243,6 +243,37 @@ points. This is more efficient, and the loss of accuracy is extremely small for
 transformations that vary smoothly between pixels. The default (``False``) is
 to use the faster option.
 
+When, for any one output pixel, the sampling region in the input image
+straddles the boundary of the input image or lies entirely outside the input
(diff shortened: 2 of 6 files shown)


astrofrog:
I think the hanging linux wheel build is unrelated

svank:
@astrofrog No worries! I've rebased this and the others. I didn't see a `CHANGES.rst`, only `CHANGES.md`, so I updated that one.

astrofrog:
Perfect, yes sorry I meant .md

astrofrog (approved):
This looks great to me!

Cadair (approved):
lgtm, but are all these CI fails expected?

astrofrog:
Oops yes missed that, I think the CI was not failing yesterday for other PRs so these might be real failures (can't check on phone)

svank:
The CI failures look unrelated, maybe a temporary failure of some server somewhere:
```
../../docs/mosaicking.rst F                                              [100%]

<snip>

027     >>> table = imagesearch('https://irsa.ipac.caltech.edu/cgi-bin/2MASS/IM/nph-im_sia?type=at&ds=asky&',
UNEXPECTED EXCEPTION: DALQueryError: _isisql_init() ../src/wsvc/2MASS/IM/imsearch.c:644
failed to connect to database tmassimg
Traceback (most recent call last):
```

That code works on my machine right now, so the failure might be resolved now. I don't know how to re-run CI though.

astrofrog:
You are right, this was intermittent - I will go ahead and merge this, there are issues with some of the wheel builds but these are unrelated to this PR.

Issue #281, opened 2022-03-14
Use asdf-astropy instead of astropy.io.misc.asdf for astropy 5.1+


pllim:
`astropy.io.misc.asdf` is deprecated. Your devdeps job is now failing because if this. See astropy/astropy#12903 and astropy/astropy#12930 .

Reported upstream at sunpy/sunpy#6026

pllim:
Hmmm... `cls = <class 'sunpy.io.special.asdf.tags.coordinates.frames.HeliographicStonyhurstCoordType'>`? cc @Cadair

Issue #270, opened 2022-01-13
TST: Unable to locate core schemas in devdeps job


pllim:
- [x] Undo #272 
- [x] Undo #285 

I added a test against dev astropy in #267 and saw these errors. Not sure how to fix. Example log: https://github.com/astropy/reproject/runs/4798998115?check_suite_focus=true

```
___________________ test_reproject_adaptive_roundtrip[asdf] ____________________

group = 'asdf.resource_mappings'
proxy_class = <class 'asdf.resource.ResourceMappingProxy'>

(very long traceback)

E       asdf.exceptions.AsdfWarning: asdf.extensions plugin from package asdf-astropy==0.1.3.dev4+ge779418 failed to load:
E       
E       AsdfWarning: asdf.resource_mappings plugin from package asdf==2.8.4.dev14+gc1f6cf9 failed to load:
E       
E       RuntimeError: Unable to locate core schemas

________________________ test_reproject_roundtrip[asdf] ________________________
(same as above)
```

PR #285, opened 2022-04-04
TST: Ignore asdf warnings in reproject tests

pllim:
~Remove asdf from~ Ignore asdf warnings in reproject tests until Sunpy can fix their stuff.

Also see:

* #270
* #281
* sunpy/sunpy#6026

### DIFF
--- a/reproject/adaptive/tests/test_core.py
+++ b/reproject/adaptive/tests/test_core.py
@@ -164,7 +164,7 @@ def test_reproject_adaptive_high_aliasing_potential():
     np.testing.assert_allclose(array_out, data_ref)
 
 
-@pytest.mark.filterwarnings('ignore:asdf.* failed to load')
+@pytest.mark.filterwarnings('ignore')
 @pytest.mark.array_compare(single_reference=True)
 @pytest.mark.parametrize('file_format', ['fits', 'asdf'])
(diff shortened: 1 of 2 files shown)


pllim:
`wheels_linux` failures (`No package geos-devel available.` for Python 3.7) is unrelated.

astrofrog:
Rather than remove the test can we xfail it?

Cadair:
Why don't we just ignore the warning?!

pllim:
If you ignore it, you will forget about it, and reproject will break spectacularly when STScI removes `astropy.io.misc.asdf` from the code base. Are you sure?

Cadair:
If we skip the test we will also forget about it, but it will never break so we will never remember :stuck_out_tongue: 

This is really a sunpy upstream issue, so best to track it and deal with it there.

pllim:
Why not move this whole test upstream so I will stop bothering you?

Cadair:
It's testing reproject and not sunpy, it just happens to be using solar data because that's what hit the bug.

I would point out that this is really *asdf*'s fault for breaking all their API so massively :p

pllim:
Okay, I can ignore the warning. But if they get dragged to the point of outright failures, I am removing them.

PR #291, opened 2022-05-21
Changed default filter kernel and boundary mode in ``reproject_adaptive``, and removed ``order`` argument.

svank:
Split from #276 

This PR changes the defaults for `reproject_adaptive` to use the Gaussian kernel, with the Hann left as an option. This is the most significant change in this PR series, as anyone using `reproject_adaptive` in their code will see very different results after updating to a version of `reproject` with this change. Their output images will be a bit blurrier, but better anti-aliased. (To be sure, every output image will also change due to the bug fixes, though those changes will hopefully be much smaller.) I'll argue that making this kernel the default is the right choice, though, because it strengthens the anti-aliasing quite a bit, and that anti-aliasing is what sets this algorithm apart from the interpolating and exact algorithms. Using the Gaussian by default offers the best anti-aliasing by default when a user comes to this function for anti-aliasing. I'm happy to discuss this point further and offer plots and examples.

As discussed, this should probably wait until after some sort of deprecation period.

EDIT: To avoid a bunch of merge conflicts, I've moved into this PR from #293 the commit that changes the default kernel. This PR now also removes the warnings we added about these future changes, removes the now-deprecated order argument, and makes notes in CHANGES.md. Should be ready to go for the next-next release.

### DIFF
--- a/docs/celestial.rst
+++ b/docs/celestial.rst
@@ -205,14 +205,14 @@ kernel---see below.)
 
 The kernel used for interpolation and averaging can be controlled with a set of
 options. The ``kernel`` argument can be set to 'hann' or 'gaussian' to set the
-function being used. The Hann window is the default, and the Gaussian window
-improves anti-aliasing and photometric accuracy (or flux conservation, when the
-flux-conserving mode is enabled) at the cost of blurring the output image by a
-few pixels. The ``kernel_width`` argument sets the width of the Gaussian
-kernel, in pixels, and is ignored for the Hann window. This width is measured
-between the Gaussian's :math:`\pm 1 \sigma` points. The default value is 1.3
-for the Gaussian, chosen to minimize blurring without compromising accuracy.
-Lower values may introduce photometric errors or leave input pixels
+function being used. The Gaussian window is the default, as it provides better
+anti-aliasing and photometric accuracy (or flux conservation, when the
+flux-conserving mode is enabled), though at the cost of blurring the output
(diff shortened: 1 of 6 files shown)


astrofrog (changes requested):
@svank - sorry for the delay in getting back to these PRs - I've now merged #276 so could you rebase this? (and definitely add an entry in ``CHANGES.rst`` for this PR)

svank:
@astrofrog What kind of warning are you thinking about? A mention in the changelog or the docs, or something like a DeprecationWarning when the code runs with the old default in effect? In any case, there is not currently a warning.

Cadair:
Yeah I would say `FutureWarning` is the most appropriate given this: https://docs.python.org/3/library/warnings.html#warning-categories

svank:
To avoid a bunch of merge conflicts, I've moved into this PR the commit that changes the default kernel. This PR now also removes the warnings we added about these future changes, removes the now-deprecated `order` argument, and makes notes in `CHANGES.md`. Should be ready to go for the next-next release.

Cadair:
@svank Incase you missed it, reproject 0.9 is out with all your PRs in.

astrofrog:
@svank - we've updated the infrastructure here to no longer require manual editing of the changelog but instead relying on PR titles, so I've made the title here more descriptive. Could you rebase this PR and remove the changes to ``CHANGES.md``?

svank:
@astrofrog Done!

astrofrog (approved):

Cadair (approved):

PR #293, opened 2022-05-23
Use "strict" boundaries by default for adaptive resampling

svank:
Split from #279

After the discussion around defaults changes, I'm separating out this commit as well. After adding a menu of boundary-handling options in #279, the old behavior can be had with the "ignore" option, and a new "strict" option gives `NaN`s for any output pixels that aren't fully determined by the input data (which generally has the effect of trimming the edges of the output image if the input image doesn't fully span the output plane). The "strict" option seems like the safest default, but I don't feel too strongly if changing this default is more work than it's worth.

Until things are merged and rebased, "Set default boundary mode to 'strict' for adaptive algo" is the only commit belonging to this PR.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+0.10 (unreleased)
+-----------------
+
+- Changed default boundary mode from ``ignore`` to ``strict`` for
+  ``reproject_adaptive``.
+
 0.9 (unreleased)
 ----------------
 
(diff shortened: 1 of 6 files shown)


astrofrog:
@svank - sorry for the delay in getting back to these PRs - I've now merged #276 so could you rebase this? (and definitely add an entry in ``CHANGES.rst`` for this PR). We can then wait until after the next release to actually merge this change.

Is there a warning yet in ``main`` about this upcoming change?

astrofrog (changes requested):
@svank - sorry for the delay in getting back to these PRs - I've now merged #276 so could you rebase this? (and definitely add an entry in ``CHANGES.rst`` for this PR). We can then wait until after the next release to actually merge this change.

Is there a warning yet in ``main`` about this upcoming change?

svank:
To avoid a bunch of merge conflicts, I'll close this PR and move this commit to the same PR that changes the default kernel.

PR #302, opened 2022-08-24
Warn about future changes to adaptive defaults

svank:
(One new commit on top of the boundary-handling PR, will rebase later)

This warns the user if the default filter kernel or boundary handling mode is being used.

```
In [3]: reproject.reproject_adaptive((data, w), w, data.shape)
<ipython-input-3-6fbe9b99ba1d>:1: FutureWarning: The default kernel will change from 'Hann' to  'Gaussian' in a future release. To suppress this warning, explicitly select a kernel with the 'kernel' argument.
  reproject.reproject_adaptive((data, w), w, data.shape)
<ipython-input-3-6fbe9b99ba1d>:1: FutureWarning: The default boundary mode will change from 'ignore' to  'strict' in a future release. To suppress this warning, explicitly select a mode with the 'boundary_mode' argument.
  reproject.reproject_adaptive((data, w), w, data.shape)
Out[3]: 
...

In [4]: reproject.reproject_adaptive((data, w), w, data.shape, kernel='hann', boundary_mode='strict')
Out[4]: 
...
```

I set the default values for these two kwargs to `'default'` so we can tell whether the user is relying on the default behavior or explicitly requesting the old behavior. I haven't mentioned the literal value of `default` in the docs and intend to stop recognizing it in the same commit that changes the defaults, and I hope this temporarily-valid value of `default` doesn't itself need a deprecation period. LMK if there's a better way.

### DIFF
--- a/reproject/adaptive/high_level.py
+++ b/reproject/adaptive/high_level.py
@@ -1,5 +1,6 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import astropy.utils
+import warnings
 
 from ..utils import parse_input_data, parse_output_projection
 from .core import _reproject_adaptive_2d
@@ -12,8 +13,9 @@ def reproject_adaptive(input_data, output_projection, shape_out=None, hdu_in=0,
                        order=None,
                        return_footprint=True, center_jacobian=False,
                        roundtrip_coords=True, conserve_flux=False,
-                       kernel='Hann', kernel_width=1.3, sample_region_width=4,
-                       boundary_mode='ignore', boundary_fill_value=0,
+                       kernel=None, kernel_width=1.3,
+                       sample_region_width=4,
+                       boundary_mode=None, boundary_fill_value=0,
                        boundary_ignore_threshold=0.5, x_cyclic=False,
                        y_cyclic=False):
     """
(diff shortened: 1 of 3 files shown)


astrofrog:
@svank - could you rebase this again? Thanks!

svank:
@astrofrog Done

reproject/adaptive/high_level.py
```diff
@@ -12,8 +13,9 @@ def reproject_adaptive(input_data, output_projection, shape_out=None, hdu_in=0,
                        order=None,
                        return_footprint=True, center_jacobian=False,
                        roundtrip_coords=True, conserve_flux=False,
-                       kernel='Hann', kernel_width=1.3, sample_region_width=4,
-                       boundary_mode='ignore', boundary_fill_value=0,
+                       kernel='default', kernel_width=1.3,
```
astrofrog, line 16: The usual way to deal with this is to actually have ``kernel=None`` rather than have ``'default'`` as a special string

astrofrog (changes requested):
Looks good but could you change 'default' to None as mentioned below?

reproject/adaptive/high_level.py
svank: Ooh, yes, I like that better. Done

astrofrog (approved):
Looks good to me assuming CI passes. Thanks! Do you have any more PRs that should be considered before I go ahead and do a release? (after which we can finalize/merge the ones to actually change the defaults). I was thinking we could perhaps plan to wait a couple of months before the next release (which would change the defaults) just to make sure we give some time for users to see the warnings.

PR #316, opened 2022-10-07
Update pinned version of Cython

astrofrog:
This may fix the Python 3.11 wheel builds

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,6 +14,10 @@ write_to = "reproject/version.py"
 skip = "cp36-* pp* *-musllinux* cp310-win32"
 test-skip = "*-macosx_arm64 *-manylinux_aarch64"
 
+[[tool.cibuildwheel.overrides]]
+select = "*-win32"
+test-requires = "scipy==1.9.1"
+
 [tool.isort]
 profile = "black"
 multi_line_output = 3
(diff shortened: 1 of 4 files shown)


astrofrog:
Looks like the main issue at the moment is the lack of wheels for scipy with Python 3.11

Cadair (approved):

pyproject.toml
```diff
@@ -14,6 +14,10 @@ write_to = "reproject/version.py"
 skip = "cp36-* pp* *-musllinux* cp310-win32"
 test-skip = "*-macosx_arm64 *-manylinux_aarch64"
 
+[[tool.cibuildwheel.overrides]]
+select = "*-win32"
+test-requires = "scipy==1.9.1"
```
Cadair, line 19: Is the fact this hasn't worked because of https://github.com/OpenAstronomy/github-actions-workflows/issues/77 ?

ConorMacBride: No, this issue is unrelated to that. You can see in the cibuildwheel logs that it does recognise the option: https://github.com/astropy/reproject/actions/runs/3328244184/jobs/5504178849#step:4:74

This issue is that cibuildwheel first installs the package before installing `test-requires`, so it’s installing the newer scipy first and failing before it has the chance to try the pinned version: https://github.com/pypa/cibuildwheel/blob/6c361f0193ca23e7355ad157a38e99903bc3c921/cibuildwheel/windows.py#L533-L543

It looks like it should work if you do `before-test = "pip install scipy==1.9.1"` instead of the `test-requires = …` line.

astrofrog: Done - thanks!

PR #330, opened 2022-12-28
Close fits files after loading

svank:
I noticed a comment pointing out that when FITS file names are passed for the input data, the file is opened but never closed. Changing that looked like a quick fix, and should be safe---after the file is opened, `parse_input_data` calls itself a few times before eventually extracting the `.data` and `.header` attributes of the selected HDU and returning those, so the HDUL object never leaves `parse_input_data`. (Not sure if the garbage collector would have eventually closed everything anyway.)

While checking that this was covered by existing tests, I saw that the `parse_input_data` tests were checking the data array that gets returned but not the corresponding WCS, so I augmented those tests to check both.

(The CI failure appears to be pre-existing)

### DIFF
--- a/reproject/tests/test_utils.py
+++ b/reproject/tests/test_utils.py
@@ -15,11 +15,18 @@ def test_parse_input_data(tmpdir):
 
     data = np.arange(200).reshape((10, 20))
 
-    hdu = fits.ImageHDU(data)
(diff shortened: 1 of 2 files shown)


astrofrog (approved):
Thanks for the fix!

PR #331, opened 2022-12-30
Fix weird quotation marks from Black auto-formatting

svank:
Noticed another little thing---a commit a few months ago auto-reformatted the code with Black. Black must have been imposing a new, larger maximum line length and it reflowed some strings that had been broken across multiple lines, but it left extra quotation marks in place while it did that

(CI failure appears to be pre-existing)


astrofrog (approved):
Thanks!

Issue #296, opened 2022-07-06
Support for common use case: reprojecting multiple images with the same original WCS 


entaylor:
I am in the midst of a very large undertaking, with many 100s of 1000s of images, each with its own variance map.

Currently i am reprojecting the image and variance maps separately.  This means i am duplicating lots of coordinate transformation computations, where in principle i could apply the pixel-to-pixel mapping from one image directly to the other.

Given that transforming images and error/variance/weight maps is a common operation, is there a straightforward way to make it possible feed multiple images into reproject?  Especially in conjunction with issue #229 this could give of order 50-75% reduction in run time for this phase of my code.

Cadair:
Just to clarify, you have the same input and output WCS for all these reproject operations or just the same input WCS?

entaylor:
Every image has an associated error map, so two arrays with the same geometry and WCS.  Then i project these to a different final WCS.  I have many images all with their own WCS, but each has its associated error map.  

Maybe it's easier in pseudo code.  What i do now is something like:
```
for (image, variance, wcs_in) in get_lots_of_data :
    img_reproj = reproject( (image, wcs_in), wcs_out)
    var_reproj = reproject( (variance, wcs_in), wcs_out)
```

What i would like to do is something like:
```
for (image, variance, wcs_in) in get_lots_of_data :
    img_reproj, var_reproj = reproject( ((image, variance), wcs_in), wcs_out)
```

The idea is just to reduce the number of repeated WCS transformations.

entaylor:
Gently bumping this for visibility.  How easy would it be to cache the pixel-to-pixel mapping to apply to multiple input images (with a common input WCS) being projected to the same output WCS?

Cadair:
One thing you could try is putting the `lru_cache` decorator on this function: https://github.com/astropy/reproject/blob/3cfc7a8f402789d338eb2379b1ba1b7b2e265738/reproject/wcs_utils.py#L15

You would probably want to tune the size of the cache (maybe to only cache the last call) so that it didn't eat a *lot* of ram. It would be good to know if it worked and then we could think about how to integrate it and make it configurable.

svank:
I'm also interested in this feature and can help with the implementation, especially for the adaptive algorithm.

[...]

PR #332, opened 2023-01-05
Add support for efficiently reprojecting multiple images with the same wcs

svank:
This addresses #296. @entaylor, want to take a look?

This PR adds support for reprojecting multiple images that share the exact same coordinate information---e.g. a data array and a corresponding error array, or an L1 and L2 image pair. By providing multiple images in one reproject call, the coordinate transformation work can be done only once before reprojecting each image, offering a speedup.

I chose the syntax of
```python
data, header = fits.getdata('data.fits', header=True)
uncertainty = fits.getdata('uncertainty.fits')
header_out = [...]

image_stack = np.stack((data, uncertainty))
reproject.reproject_adaptive((image_stack, header), header_out)
```
i.e. multiple images passed as an extra input array dimension. There's no support for doing this by passing some sort of mult-HDU FITS files directly to the reproject function---it seems like it would be too much work to verify each HDU has the same WCS.

I tried to mimic the way numpy handles broadcasting operations between two arrays---the trailing dimensions are matched up, while extra leading dimensions are looped over. For reproject, I'm taking the trailing dimensions of the input and output arrays to correspond to the WCS dimensions, and any extra leading dimensions beyond what the WCS contains are looped over. Any number of extra leading dimensions is supported---maybe a user with data and uncertainty arrays from both L1 and L2 wants to do a `2 × 2 × nx × ny` input array.

When `shape_out` is provided, it can contain the extra dimensions that the input array contains (and they must match exactly), or `shape_out` can be just the shape of a single output image, and the extra dimensions are prepended automatically.

The implementation is a bit different for each of the interpolation, adaptive, and exact algorithms:

For interpolation, I'm computing the pixel-to-pixel mapping once, and then looping over the extra dimensions, calling `map_coordinates` once for each image. To support the new blocked mode (which is only in this algorithm), I did change how that mode approaches 3+ dimensional WCSes. It breaks up the data in only two dimensions---those were originally the first two dimensions, but I changed it to be the last two dimensions, since those will always be WCS dimensions. (The alternatives would be "the first two WCS dimensions, which are not necessarily the first two array dimensions" or "the first two array dimensions, which are not necessarily WCS dimensions", which would both complicate the logic a bit.)

For adaptive, I made the changes at the Cython level, since that's where some of the coordinate work occurs. The few spots that update the output array (the inner-most part of the loop) are modified to do everything along the extra dimension.

For exact, the changes are at the Python level, similar to interpolation with a loop over the extra dimensions after calculating the coordinates once. This algorithm has a parallel-processing mode which could in principle parallelize across the extra dimensions, but then there would have to be separate parallel modes depending on whether there are extra dimensions, so I've left the parallelization inside the loop over extra dimensions.

My speed test results with two PSP/WISPR images, each about 1k x 1k:
```
>>> two_images = np.stack((d1, d2))

>>> %%timeit
... dout = reproject.reproject_interp((two_images, w1), wout, d1.shape, return_footprint=False)
831 ms ± 8.38 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

>>> %%timeit
... dout1 = reproject.reproject_interp((d1, w1), wout, d1.shape, return_footprint=False)
... dout2 = reproject.reproject_interp((d2, w1), wout, d1.shape, return_footprint=False)
1.75 s ± 21.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


>>> %%time
... dout = reproject.reproject_adaptive((two_images, w1), wout, d1.shape, return_footprint=False)
1.58 s ± 23.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

>>> %%timeit
... dout1 = reproject.reproject_adaptive((d1, w1), wout, d1.shape, return_footprint=False)
... dout2 = reproject.reproject_adaptive((d2, w1), wout, d1.shape, return_footprint=False)
3.27 s ± 45.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


>>> %%timeit
... dout = reproject.reproject_exact((two_images, w1), wout, d1.shape, return_footprint=False)
19 s ± 961 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

>>> %%timeit
... dout1 = reproject.reproject_exact((d1, w1), wout, d1.shape, return_footprint=False)
... dout2 = reproject.reproject_exact((d2, w1), wout, d1.shape, return_footprint=False)
19.7 s ± 733 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
tl;dr A 2x speedup for two images in interpolating or adaptive reprojection

(There's one pre-existing test failure that's gonna tank CI.)

### DIFF
--- a/docs/celestial.rst
+++ b/docs/celestial.rst
@@ -419,3 +419,51 @@ Or if you're dealing with FITS files, you can skip the numpy memmap step and use
     ...                                   return_footprint=False,
     ...                                   independent_celestial_slices=True)
     >>> outhdu.flush()
+
+Multiple images with the same coordinates
+=========================================
+
+If you have multiple images with the exact same coordinate system (e.g. a raw
+image and a corresponding processed image) and want to reproject both to the
+same output frame, it is faster to compute the coordinate mapping between input
+and output pixels only once and re-use this mapping for each reprojection. This
+is supported by passing multiple input images as an additional dimension in the
+input data array. When the input array contains more dimensions than the input
+WCS describes, the extra leading dimensions are taken to represent separate
+images with the same coordinates, and the reprojection loops over those
+dimensions after computing the pixel mapping. For example:
+
+.. doctest-skip::
+    >>> raw_image, header_in = fits.getdata('raw_image.fits', header=True)
+    >>> bg_subtracted_image = fits.getdata('background_subtracted_image.fits')
+    >>> header_out = # Prepare your desired output projection here
+    >>> # Combine the two images into one array
+    >>> image_stack = np.stack((raw_image, bg_subtracted_image))
+    >>> # We provide a header that describes 2 WCS dimensions, but our input
+    >>> # array shape is (2, ny, nx)---the 'extra' first dimension represents
+    >>> # separate images sharing the same coordinates.
+    >>> reprojected, footprint = reproject.reproject_adaptive(
+    ...         (image_stack, header_in), header_out)
+    >>> # The shape of `reprojected` is (2, ny', nx')
+    >>> reprojected_raw, reprojected_bg_subtracted = reprojected[0], reprojected[1]
+
+For :func:`~reproject.reproject_interp` and
+:func:`~reproject.reproject_adaptive`, this is approximately twice as fast as
+reprojecting the two images separately. For :func:`~reproject.reproject_exact`
+the savings are much less significant, as producing the coordinate mapping is a
+much smaller portion of the total runtime for this algorithm.
+
+When the output coordinates are provided as a WCS and a ``shape_out`` tuple,
+``shape_out`` may describe the output shape of a single image, in which case
+the extra leading dimensions are prepended automatically, or it may include the
+extra dimensions, in which case the size of the extra dimensions must match
+those of the input data exactly.
+
+While the reproject functions can accept the name of a FITS file as input, from
+which the input data and coordinates are loaded automatically, this multi-image
+reprojection feature does not support loading multiple images automatically
+from multiple HDUs within one FITS file, as it would be difficult to verify
+automatically that the HDUs contain the same exact coordinates. If multiple
+HDUs do share coordinates and are to be reprojected together, they must be
+separately loaded and combined into a single input array (e.g. using
+``np.stack`` as in the above example).
--- a/reproject/adaptive/core.py
+++ b/reproject/adaptive/core.py
@@ -104,19 +104,50 @@ def _reproject_adaptive_2d(
     to numerical issues, causing a band of NaN values around the edge of the
     final image (when reprojecting from helioprojective to heliographic).
     See https://github.com/astropy/reproject/issues/195 for more information.
+
+    Note
+    ----
+    If the input array contains extra dimensions beyond what the input WCS has,
+    the extra leading dimensions are assumed to represent multiple images with
+    the same coordinate information. The transformation is computed once and
+    "broadcast" across those images.
     """
 
     # Make sure image is floating point
     array_in = np.asarray(array, dtype=float)
+    shape_out = tuple(shape_out)
 
     # Check dimensionality of WCS and shape_out
     if wcs_in.low_level_wcs.pixel_n_dim != wcs_out.low_level_wcs.pixel_n_dim:
         raise ValueError("Number of dimensions between input and output WCS should match")
-    elif len(shape_out) != wcs_out.low_level_wcs.pixel_n_dim:
-        raise ValueError("Length of shape_out should match number of dimensions in wcs_out")
+    elif len(shape_out) < wcs_out.low_level_wcs.pixel_n_dim:
+        raise ValueError("Length of shape_out too small for number of dimensions in wcs_out")
+    elif len(array_in.shape) < wcs_in.low_level_wcs.pixel_n_dim:
+        raise ValueError("Too few input-data dimensions for number of dimensions in wcs_out")
+    elif len(array_in.shape) != len(shape_out):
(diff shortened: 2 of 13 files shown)


astrofrog (commented):
This is great @svank, thank you!

I agree with putting the onus on the user to construct the stacked array otherwise we would indeed need to somehow compare WCSes etc. However I wonder if you could put an example in the docs showing how to do this?

Once this is done I would be happy to merge. I'll wait a few days anyway in case @entaylor wants to try this out.

@svank - please ping me once you've added the docs!

astrofrog:
@svank - when you work on this next, could you rebase? (the CI issues should be fixed)

svank:
@astrofrog Done! Thanks for reviewing!

astrofrog:
Thanks!

After a quick discussion with @Cadair, I wanted to bring up a point related to the API - at the moment, this PR might tie us in to having leading dimensions being the ones to ignore, but there are some other use cases we might want to consider. For example, one might want to pass a 3D spectral+celestial cube, and reprojecting just the celestial or spectral part - in this case, we might want to have a way to specify the input WCS as being 3D and then giving the part of the WCS we want to reproject as the target WCS and then having reproject figure out internally which dimensions to ignore/loop over. This might not be in conflict with what is done in this PR because in this PR I think you are saying that leading dimensions are ignored compared to the dimensionality of the *input* WCS. However I wonder if we should perhaps allow the dimensions to be ignored to be configurable (defaulting to extra leading dimensions). Although again maybe as long as we know we can add that option in future maybe what is here is fine. At least I think we should make it clear that what matters here is dimensionality of the data compared to the dimensionality of the input WCS.

svank:
Glad the API choice is getting some review. I think you're right that if we're comfortable defaulting to taking the leading dimensions as the extra loop dimensions (and that feels sensible to me), this PR doesn't preclude adding additional flexibility later. Right now, the input and output WCS are required to have the same number of dimensions---this was already explicitly checked for in [`interpolation/core.py`](https://github.com/astropy/reproject/pull/332/files#diff-d13120906e2360e5122847bbdad1f2618f60230b763434d038ccdc537d407a6dL12), [`adaptive/core.py`](https://github.com/astropy/reproject/pull/332/files#diff-d5e34c5db9c6782c05cb9351148ff56d674300289367efde2a28d7a5bf711b15L113), and [`spherical_intersect/high_level.py`](https://github.com/astropy/reproject/pull/332/files#diff-b59ea65003c8b798b772558f3803493e3dbf74e297e1f1ee89d44e42cfc0f3ccL64)---so there's no practical distinction on which WCS we're comparing the input array shape to, but we could document it as the input WCS to offer flexibility later.

What I'm interpreting as the future API you're envisioning is matching up the output WCS dimensions to input WCS dimensions to determine the transformation and looping over any other input dimensions, whether 'extra' dimensions not in the input WCS or input WCS dimensions not present in the output WCS, and maybe with some sort of `wcs_to_array_axis` argument that specifies the input array dimension corresponding to each input WCS dimension and which defaults to mapping the WCS dimensions in the normal order to the last array dimensions. If that's indeed what you're imagining, I think that's out-of-scope for this PR but builds naturally on this PR, with the right future default behavior being the current behavior. The things enabled by those future changes should still be achievable with this PR through suitable edits by the user to the input and output WCSes and array dimensions, so that flexibility would be about convenience. I think this PR also creates a natural point in each algorithm at which to implement that flexibility---for each one there's a spot where I adjust the input array to always have exactly one leading loop dimension (of size one if no looping is actually needed), and then I impose the correct output shape at the end, so those would be good places to do more conversions between what the user provided and some standardized axis order the core algorithms can work with.

So if I make sure the docs and code refer specifically to the input WCS dimensionality, do you think this is in a good current state?

astrofrog:
I agree with your assessment and indeed the PR will be ready to go after those tweaks.

entaylor:
@svank : Thank you for this!  I will be very excited to give this a spin when i come back from leave early next week.

The API question is worth considering carefully.  Indeed, IFS data is one potentially very interesting use case for this functionality.  

From my perspective, for what it's worth: Typically, IFS data will have 3 WCS dimensions; two spatial and one spectral.  If one wanted to reproject IFS data, then i think the way this could be done with the current structure would be to pass the 3D IFS data array, and then with only two of the three WCS dimensions.  And if i wanted to reproject both IFS data and weight/variance arrays, then i would pass in a 4D array; e.g. of shape (2, nwl, nx, ny).  

Does that make sense from your point of view?  And/or does it impact at all the structure that you are envisaging?

svank:
Yes, you have it right. With this PR you can remove the spectral axis from the WCS and ensure the spectral array axis is first in order to do a spatial resampling with optimized looping (i.e. treat the xy slice at every lambda value as a separate image, but all with the same coordinates), and you can also add in a data/variance axis and pass in a (2, nwl, nx, ny) array, with looping over the first two dimensions. The future API goals would make this more convenient, and I think the IFS case fits well.

IFS data is indeed a really compelling use case for this optimization, with so many slices with identical x/y coordinates. In some quick testing, I'm seeing a 20-25x speedup for what I assume is a typical IFS data cube size:
image

svank:
Docs should now specifically refer to the input WCS

astrofrog (approved):
Thanks, happy to approve this now!

However I will wait for @entaylor to try this out early next week before going ahead and merging in case there are any issues.

astrofrog:
@entaylor - did you get a chance to look at this?

astrofrog:
I am going to go ahead and merge this as I am going to be preparing a new release shortly - but @entaylor if you get to try this out and run into any issues please post a comment here!

PR #336, opened 2023-02-15
Added min and median combine_functions

minzastro:
I add two new combine functions. One is simply 'min' - very similar to what exists already.
The other one is 'median', which required some per-block processing. This is probably slower - I made no tests for performance yet.

### DIFF
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -1,14 +1,14 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 
+import warnings
+
 import numpy as np
 
 from ..utils import parse_input_data, parse_input_weights, parse_output_projection
 from .background import determine_offset_matrix, solve_corrections_sgd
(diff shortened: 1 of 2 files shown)


astrofrog:
Thanks for the pull request! I'll try and take a look at this soon! In the meantime, I noticed that some of the code has been re-formatted - could you run ``tox -e codestyle`` to format it back and commit the changes?

astrofrog:
@minzastro - sorry for not getting back to this. Now that we use dask internally in a few places in reproject I think we should also consider using dask to handle the median calculation to avoid having to do all the block handling ourselves. I'll open an alternative PR to try this out.

astrofrog:
I'm running into issues using dask, have posted a question here: https://dask.discourse.group/t/constructing-a-sparse-dask-array-from-numpy-arrays/2187 - once there is a reply, we can decide how to proceed.

astrofrog:
Done in #622 (for median) and earlier PRs for min. I will close this now, but thank you for the PR!

Issue #309, opened 2022-09-09
Make `shape_out` optional if input WCS has `array_shape`


Cadair:
In APE 14 there is a nicely specified way of getting the array shape, we should support reading it.

PR #345, opened 2023-03-03
Add support for specifying output projection as APE 14 WCS with array_shape defined

astrofrog:
This also refactors output_projection tests to use a fixture.

Fixes https://github.com/astropy/reproject/issues/309

### DIFF
--- a/reproject/utils.py
+++ b/reproject/utils.py
@@ -9,7 +9,7 @@ import numpy as np
 from astropy.io import fits
 from astropy.io.fits import CompImageHDU, HDUList, Header, ImageHDU, PrimaryHDU
 from astropy.wcs import WCS
-from astropy.wcs.wcsapi import BaseLowLevelWCS, SlicedLowLevelWCS
+from astropy.wcs.wcsapi import BaseLowLevelWCS, BaseHighLevelWCS, SlicedLowLevelWCS
 from astropy.wcs.wcsapi.high_level_wcs_wrapper import HighLevelWCSWrapper
 from dask.utils import SerializableLock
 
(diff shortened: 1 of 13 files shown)


reproject/utils.py
```diff
@@ -142,9 +142,11 @@ def parse_output_projection(output_projection, shape_in=None, shape_out=None, ou
                     "Need to specify shape since output header "
                     "does not contain complete shape information"
                 )
-    elif isinstance(output_projection, BaseHighLevelWCS):
+    elif isinstance(output_projection, (BaseLowLevelWCS, BaseHighLevelWCS)):
```
Cadair, line 145: oh man this change is needed, been meaning to raise an issue for this for ages.

reproject/utils.py
```diff
@@ -142,9 +142,11 @@ def parse_output_projection(output_projection, shape_in=None, shape_out=None, ou
...
         wcs_out = output_projection
-        if shape_out is None:
+        if getattr(wcs_out, "array_shape") is not None:
+            shape_out = wcs_out.array_shape
```
Cadair, line 148: The issue is that not all `HighLevelWCS` classes also implement the low level API (notably the wrapper doesn't). So I think you might not be able to get to `array_shape` if the `output_projection` is an instance of wrapper. (You would have to go through `low_level_wcs`.

It might be worth standardising on low level?

astrofrog: We've been allowing high level WCS for a little while so I think we just have to support both (in fact we rely on the high level API internally). I've expanded the test suite and made sure things work for both now.

PR #351, opened 2023-03-13
Generalize reproject_and_coadd for N-dimensional data, and add option to specify blank pixel value and progress bar

keflavich:
This enables memmap'd output, which will be needed for very large (greater-than-memory) output cubes

EDIT: it also extends `reproject_and_coadd` into three dimensions

### DIFF
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -195,8 +195,9 @@ def reproject_and_coadd(
             xc = np.array([-0.5, nx - 0.5, nx - 0.5, -0.5])
             yc = np.array([-0.5, -0.5, ny - 0.5, ny - 0.5])
             zc = np.array([-0.5, nz - 0.5])
-            xc_out, yc_out = wcs_out.celestial.world_to_pixel(wcs_in.celestial.pixel_to_world(xc, yc))
-            zc_out = wcs_out.spectral.world_to_pixel(wcs_in.spectral.pixel_to_world(zc))
+            # TODO: figure out what to do here if the low_level_wcs doesn't  support subsetting
+            xc_out, yc_out = wcs_out.low_level_wcs.celestial.world_to_pixel(wcs_in.celestial.pixel_to_world(xc, yc))
+            zc_out = wcs_out.low_level_wcs.spectral.world_to_pixel(wcs_in.spectral.pixel_to_world(zc))
             shape_out_cel = shape_out[1:]
 
         # Determine the cutout parameters
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -11,6 +11,10 @@ from .subset_array import ReprojectedArraySubset
 __all__ = ["reproject_and_coadd"]
 
 
+def _noop(iterable):
+    return iterable
+
+
 def reproject_and_coadd(
     input_data,
     output_projection,
@@ -25,7 +29,7 @@ def reproject_and_coadd(
     output_array=None,
     output_footprint=None,
     block_sizes=None,
-    progressbar=False,
+    progress_bar=None,
     blank_pixel_value=np.nan,
     **kwargs,
 ):
@@ -80,7 +84,7 @@ def reproject_and_coadd(
         `~astropy.io.fits.HDUList` instance, specifies the HDU to use.
     reproject_function : callable
         The function to use for the reprojection.
-    combine_function : { 'mean', 'sum', 'median', 'first', 'last', 'min', 'max' }
+    combine_function : { 'mean', 'sum', 'first', 'last', 'min', 'max' }
         The type of function to use for combining the values into the final
         image. For 'first' and 'last', respectively, the reprojected images are
         simply overlaid on top of each other. With respect to the order of the
@@ -103,12 +107,15 @@ def reproject_and_coadd(
         specified with `shape_out` or derived from the output projection.
     block_sizes : list of tuples or None
         The block size to use for each dataset.  Could also be a single tuple
-        if you want the sample block size for all data sets
-    progressbar : False
-        If specified, use this as a progressbar to track loop iterations over
+        if you want the sample block size for all data sets.
+    progress_bar : callable, optional
+        If specified, use this as a progress_bar to track loop iterations over
         data sets.
+    blank_pixel_value : float, optional
+        Value to use for areas of the resulting mosaic that do not have input
+        data.
 
-    kwargs
+    **kwargs
         Keyword arguments to be passed to the reprojection function.
 
     Returns
@@ -127,19 +134,16 @@ def reproject_and_coadd(
 
     # Validate inputs
 
-    if combine_function not in ("mean", "sum", "median", "first", "last", "min", "max"):
-        raise ValueError("combine_function should be one of mean/sum/median/first/last/min/max")
-    elif combine_function == "median":
-        # Note to devs: the exception shoudl be raised as early as possible
-        raise NotImplementedError("combine_function='median' is not yet implemented")
+    if combine_function not in ("mean", "sum", "first", "last", "min", "max"):
+        raise ValueError("combine_function should be one of mean/sum/first/last/min/max")
 
     if reproject_function is None:
         raise ValueError(
             "reprojection function should be specified with the reproject_function argument"
         )
 
-    if not progressbar:
-        progressbar = lambda x: x
+    if progress_bar is None:
+        progress_bar = _noop
 
     # Parse the output projection to avoid having to do it for each
 
@@ -165,7 +169,7 @@ def reproject_and_coadd(
     if match_background:
         arrays = []
 
-    for idata in progressbar(range(len(input_data))):
+    for idata in progress_bar(range(len(input_data))):
         # We need to pre-parse the data here since we need to figure out how to
         # optimize/minimize the size of each output tile (see below).
         array_in, wcs_in = parse_input_data(input_data[idata], hdu_in=hdu_in)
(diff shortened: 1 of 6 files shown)


reproject/mosaicking/coadd.py
```diff
@@ -22,6 +22,9 @@ def reproject_and_coadd(
     combine_function="mean",
     match_background=False,
     background_reference=None,
+    final_array=None,
+    final_footprint=None,
```
astrofrog, line 26: Maybe it would make sense to simply use ``output_array`` and ``output_footprint`` for consistency with the individual reproject functions.

reproject/mosaicking/coadd.py
```diff
@@ -82,6 +85,18 @@ def reproject_and_coadd(
...
+        The final output array.  Specify this if you already have an
+        appropriately-shaped array to store the data in.  Must match shape
+        specified with `shape_out` or derived from the output
+        projection.
+    final_footprint : array or None
+        The final output footprint array.  Specify this if you already have an
+        appropriately-shaped array to store the data in.  Must match shape
+        specified with `shape_out` or derived from the output projection.
+    block_sizes : list of tuples or None
+        The block size to use for each cube (optional; meant for dask use)
```
astrofrog, line 98: Don't mention 'cube' here, maybe just input dataset? And don't need to mention dask here, block sizes can be used even with numpy input. You can just say that these are passed to the ``block_size`` argument of the individual reproject functions.

reproject/mosaicking/coadd.py
```diff
@@ -106,6 +121,14 @@ def reproject_and_coadd(
 
     wcs_out, shape_out = parse_output_projection(output_projection, shape_out=shape_out)
 
+    if final_array is not None and final_array.shape != shape_out:
+        raise ValueError("If you specify an output array, it must have a shape matching "
+                         "the output shape")
+    if final_footprint is not None and final_footprint.shape != shape_out:
+        raise ValueError("If you specify an output footprint array, it must have a shape matching "
+                         "the output shape")
```
astrofrog, line 129: Could you re-write these error messages to mention the expected and actual shape in each case?

reproject/mosaicking/coadd.py
```diff
@@ -165,6 +188,9 @@ def reproject_and_coadd(
 
         shape_out_indiv = (jmax - jmin, imax - imin)
 
+        if block_sizes is not None:
+            kwargs['block_size'] = block_sizes[idata]
```
astrofrog, line 192: It would be nice to allow ``block_sizes`` to perhaps be a single shape that would be the same for all images. To some extent, in the dask case (once implemented) the block size won't need to be passed in as it will be inferred from the dask arrays themselves so at that point maybe the single block size input would actually become the most commonly used.

astrofrog (changes requested):
This is useful, thanks! Just some comments below, and be sure to include some tests - thanks!

reproject/mosaicking/coadd.py
keflavich: I like that idea; `final_array` was just an internal use anyway.

reproject/mosaicking/coadd.py
keflavich: right, this might be good to refactor but I'll make it flexible for now

reproject/mosaicking/coadd.py
```diff
@@ -154,10 +154,20 @@ def reproject_and_coadd(
...
+            yc = np.array([-0.5, -0.5, ny - 0.5, ny - 0.5])
+            xc_out, yc_out = wcs_out.world_to_pixel(wcs_in.pixel_to_world(xc, yc))
+            shape_out_cel = shape_out
+        elif array_in.ndim == 3:
+            nz, ny, nx = array_in.shape
+            xc = np.array([-0.5, nx - 0.5, nx - 0.5, -0.5])
+            yc = np.array([-0.5, -0.5, ny - 0.5, ny - 0.5])
+            zc = np.array([-0.5, nz - 0.5])
+            xc_out, yc_out = wcs_out.celestial.world_to_pixel(wcs_in.celestial.pixel_to_world(xc, yc))
+            zc_out = wcs_out.spectral.world_to_pixel(wcs_in.spectral.pixel_to_world(zc))
```
astrofrog, line 170: Are the previous two lines done separately for memory efficiency?

keflavich: no, for readability - it's just easier for me to see this if they're separated into chunks.  Otherwise I think we need a full 2x3x3 grid with 0's appended on each side, which is going to be a lot of lines of code that don't do anything

reproject/mosaicking/coadd.py
keflavich: done

reproject/mosaicking/coadd.py
keflavich: done

reproject/mosaicking/coadd.py
```diff
@@ -144,26 +178,43 @@ def reproject_and_coadd(
...
+        elif array_in.ndim == 3:
+            kmin = max(0, int(np.floor(zc_out.min() + 0.5)))
+            kmax = min(shape_out[0], int(np.ceil(zc_out.max() + 0.5)))
+            if isinstance(wcs_out, WCS):
+                wcs_out_indiv = wcs_out[kmin:kmax, jmin:jmax, imin:imax]
+            else:
+                wcs_out_indiv = SlicedLowLevelWCS(
+                    wcs_out.low_level_wcs, (slice(kmin, kmax), slice(jmin, jmax), slice(imin, imax))
+                )
+            shape_out_indiv = (kmax - kmin, jmax - jmin, imax - imin)
```
keflavich, line 210: I've gotten _something_ wrong:
```
ipdb> array_in.shape
(6, 80, 101)
ipdb> shape_out_indiv
(4, 161, 152)
```

keflavich: no that's ok, nvm

keflavich:
I'm a little stymied on this:
`ValueError: Chunks do not add up to shape. Got chunks=((4,), (4, 100, 101), (4, 100, 101)), shape=(4, 174, 173)`
It looks like the block_size isn't playing nice with the different-shaped output; I think that relies on your WIP PR @astrofrog ?

keflavich:
@astrofrog I have some questions now:

The loop for `reproject_and_coadd` is presently _not_ lazy or possible to make lazy, correct?  We need your daskified PR to make it lazy?

I think I'm going to move the write-to-big-array into the parent loop, because otherwise everything still has to be held in memory

keflavich:
ahhh, there's good reason for that - background matching!  Yikes.

keflavich:
so, as is, this does coadding on a dataset-by-dataset basis - which, for cubes, means a cube-by-cube basis.  It may be more efficient to do it on a plane-by-plane basis.  That seems like a lot of additional refactoring work though

keflavich:
There is a huge efficiency boost to be gained if we can reproject directly into the output array, _but_ that results in _overwriting_ the individual values instead of adding to them.

keflavich:
@astrofrog I don't see any way to do this, but do you have ideas?  Basically, instead of `map_coordinates(..., output=array)` writing directly into memory, it would do `output += map_coordinates(...)`.  That looks like it is problematic at the lowest levels, and I don't know if it would work at all for other reprojection methods.

This sacrifices flexibility even further to improve speed and robustness and reduce memory use.  In the case where you don't want to match backgrounds, this approach just loads the unweighted data into an array and the footprint into another array and divides them at the end.  The robustness bit is that we can flush to disk granularly.

I'm not sure this is worth pursuing further, though, because of the depth of refactor that would be needed.  Maybe this is something just solved well enough by dask.

astrofrog:
I'll have a think about all this! Should be able to work on it a bit this week.

keflavich:
@astrofrog I've been using this PR in practice for a while now,  and I just had a look into rebasing, but the codebase has diverged a ton, making this a rather challenging rebase.  I'd like to go ahead with it, but this time make sure the changes can get merged.  What's your feeling - is there anything blocking this if it's rebased?

keflavich:
ok the rebase wasn't as bad as I thought, but there were some confusing items that I am not sure I've resolved yet.

astrofrog:
I can try and prioritise this!

keflavich:
with the new modes (first, last, min, max) we again face duplicating code because we need a version that does, and another version that does not, try to background-match before the final combination step.  The current approach simply breaks if match_background is not on and one tries first/last/min/max.  Is there a more elegant approach?

astrofrog:
Will investigate!

astrofrog:
To try and keep things a bit simpler I've extracted out the changes related to specifying output arrays/footprints into a separate PR: https://github.com/astropy/reproject/pull/387

astrofrog:
In terms of supporting 3D arrays (and also the median combine mode), I think things are getting complicated enough that it's worth thinking about a different approach for doing the co-adding. I'm working on an experimental re-write of ``reproject_and_coadd`` which uses dask internally and will open an alternative PR soon so that we can compare and see what works best in practice.

astrofrog:
See https://github.com/astropy/reproject/pull/388 for an alternative.

keflavich:
@astrofrog I'm still using this branch in production.  It's apparently working (verification in progress).  I clearly ran out of time to test 388.  Any further thoughts on which direction to take?  Both PRs are big, I don't immediately remember/know what the reasons are for one or the other.

astrofrog:
Sorry for dropping the ball on this, I'll try and see if I can wrap things up in the next week or so.

keflavich:
my last commit adds a hack to solve this issue: https://github.com/radio-astro-tools/spectral-cube/issues/900 (should've posted that here, perhaps, but it isn't obvious to me where it came from)

keflavich:
@astrofrog I'm at a workshop with @e-koch, and I'd like to show off that we can do the mosaicing in https://github.com/radio-astro-tools/spectral-cube/pull/868.  Any chance we can make a little progress on this shortly?

astrofrog:
@keflavich yes sorry for the delay, I will be back at work on Monday and will try and prioritise wrapping this up.

astrofrog (approved):
I pushed a couple of commits cleaning this up - I've generalized it to N dimensions which simplifies things, and changed the default for blank_pixel_value to 0 for backward-compatibility. Thanks @keflavich, and sorry for taking so long to wrap this up!

PR #352, opened 2023-03-16
Started adding support for allowing dask arrays as input

astrofrog:
This PR adds support for specifying the input data as a dask array - this covers only the general case mentioned in  https://github.com/astropy/reproject/issues/362 of **Input where all dimensions are to be reprojected**

As mentioned in #362, the general case requires us to write out the dask array to disk and reload it as a Numpy memmap array. There will be ways of making common use cases more efficient as outlined in #362 but for now this PR implements the general approach.

### DIFF
--- a/reproject/interpolation/high_level.py
+++ b/reproject/interpolation/high_level.py
@@ -1,8 +1,8 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import os
 
-from dask import array as da
 from astropy.utils import deprecated_renamed_argument
+from dask import array as da
 
 from ..utils import _reproject_blocked, parse_input_data, parse_output_projection
 from .core import _reproject_full
--- a/reproject/interpolation/high_level.py
+++ b/reproject/interpolation/high_level.py
@@ -2,7 +2,6 @@
 import os
 
 from astropy.utils import deprecated_renamed_argument
-from dask import array as da
 
 from ..utils import _reproject_blocked, parse_input_data, parse_output_projection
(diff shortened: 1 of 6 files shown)


reproject/utils.py
```diff
@@ -258,18 +258,41 @@ def _reproject_blocked(
...
+            array_in.to_zarr(filename)
+            array_in = da.from_zarr(filename)
+            # Then store this in the Numpy memmap array (schedulers other than
+            # synchronous don't work well when writing to a Numpy array)
+            da.store(
+                array_in,
+                array_in_memmapped,
+                compute=True,
+                scheduler="synchronous",
+            )
```
astrofrog, line 292: Note to self: should probably remove temporary files as we go as soon as they aren't needed otherwise /tmp might fill up

astrofrog: This is also a limitation of the approach so far that we actually have to write the dask array to disk. However, I don't see a way around this until there is a version of map_coordinates that works nicely with dask arrays, at least as far as the interpolation reprojection is concerned.

reproject/interpolation/high_level.py
```diff
@@ -116,6 +117,9 @@ def reproject_interp(
     if isinstance(order, str):
         order = ORDER[order]
 
+    if isinstance(array_in, da.core.Array) and block_size is None:
+        block_size = array_in.chunksize
```
astrofrog, line 121: Not right - we shouldn't need to care about the input chunk size.

PR #361, opened 2023-05-18
If shape_out is specified, use this over the array_shape attribute of a WCS object

astrofrog:
Still needs a regression test


astrofrog:
This is pretty straightforward so merging!

PR #365, opened 2023-05-22
Fix support for NDData objects with dask .data attributes

astrofrog:
Also remove some unused code


astrofrog:
(I don't think it will make much difference for your images)

PR #366, opened 2023-05-25
Add despike_jacobian option for adaptive resampling

### LINKED ISSUE #195: Fix edge effects in reproject_adaptive
The reproject_adaptive function in https://github.com/astropy/reproject/pull/52 (which uses the DeForest 2004 algorithm) can run into issues with all-sky maps, due to the way the derivatives are computed. Currently, these are computed by computing the coordinate transform for coordinates that go from -0.5 to shape - 0.5 - however, coordinates that lie exactly on the edge of a map may be subject to numerical issues, causing a band of NaN values around the edge of the final image (when reprojecting from helioprojective to heliographic). I can upload an example once #52 is merged.

svank:
This adds an option to smooth over numerical quirkiness when, for example, the longitude wraps around in an all-sky map. The quirk occurs when computing the Jacobian, which is d(input-image pixel) / d(output-image pixel). This is done by finite differences, so when the input coordinates jump from one side of the map to the other, the Jacobian comes out extremely large, despite the actual transformation being unremarkable at that location. The new option looks for outlier Jacobian values, measured relative to a 3x3 local neighborhood, and replaces those outliers with a mean of the non-outlier values in the neighborhood.

Before this PR, the extremely-large Jacobian would cause the input-image region that is to be sampled to be extremely large---larger than a safety-valve threshold in the code that just outputs a `nan` for that output pixel. This caused a band of `nans` near the wrap-point's location in the output image. With this change, that seam is handled correctly.

Looks like this might close #195.

```python
starfield_sample, _ = reproject.reproject_adaptive(
    (all_sky_map, all_sky_wcs), data_wcs, data.shape,
    x_cyclic=True, despike_jacobian=False)
starfield_sample_despiked, _ = reproject.reproject_adaptive(
    (all_sky_map, all_sky_wcs), data_wcs, data.shape,
    x_cyclic=True, despike_jacobian=True)

plt.figure(figsize=(10,5))
plt.subplot(121, projection=data_wcs)
plt.imshow(np.isnan(starfield_sample))
plt.title("np.isnan(starfield_sample)")

plt.subplot(122, projection=data_wcs)
plt.imshow(np.isnan(starfield_sample_despiked))
plt.title("np.isnan(starfield_sample_despiked)")
```
image

### DIFF
--- a/docs/celestial.rst
+++ b/docs/celestial.rst
@@ -243,6 +243,20 @@ points. This is more efficient, and the loss of accuracy is extremely small for
 transformations that vary smoothly between pixels. The default (``False``) is
 to use the faster option.
 
+In some situations (e.g. an all-sky map, with a wrap point in the longitude),
+extremely large Jacobian values may be computed which are artifacts of the
+coordinate system definition, rather than reflecting the actual nature of the
+coordinate transformation. This may result in a band of ``nan`` pixels in the
+output image. In these situations, if the actual transformation is
+approximately constant in the region of these artifacts, the
+``despike_jacobian`` option should be enabled. If enabled, the typical
+magnitude (distance from the determinant) of the Jacobian matrix, ``Jmag2 =
+sum_j sum_i (J_ij**2)``, is computed for each pixel and compared to the 25th
+percentile of that value in the local 3x3 neighborhood (i.e. the third-lowest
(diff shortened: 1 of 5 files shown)


astrofrog (approved):
Looks great, thanks!

PR #374, opened 2023-07-13
Refactor blocked/parallel reprojection

astrofrog:
This refactors the handling of blocked/parallel reprojection with dask following a conversation with @Cadair.

* [x] Ensure that we clean up all temporary directories with Numpy memmaps and zarr arrays
* [x] Add a ``return_type`` option to ``reproject_interp`` which can be ``'numpy'`` (default) or ``'dask'``. When the latter is used, nothing is computed when ``reproject_interp`` is called so this is very fast
* [x] Document that to use schedulers other than synchronous or processes, users should set ``return_type`` to ``'dask'`` and then compute it with the desired scheduler (e.g. a dask distributed scheduler). However, we need to make it very clear that the code is not thread-safe, so that multi-threaded schedulers can't be used (I don't think there is a way for us to check that)
* [x] Expand the blocked/parallel reprojection to the adaptive and spherical intersect algorithms
* [x] Make sure we properly test ``roundtrip_coords`` and ``order`` - these were not getting passed through to the blocked/parallel reprojection so were getting ignored, but need to add tests to make sure these options do something

### DIFF
--- a/docs/celestial.rst
+++ b/docs/celestial.rst
@@ -385,12 +385,9 @@ reprojection using the spherical polygon intersection of input and output pixels
 
     >>> from reproject import reproject_exact
 
-In addition to the arguments described in :ref:`common`, an optional
-``parallel=`` option can be used to control whether to parallelize the
-reprojection, and if so how many cores to use (see
-:func:`~reproject.reproject_exact` for more details). For this algorithm, the
-footprint array returned gives the exact fractional overlap of new pixels with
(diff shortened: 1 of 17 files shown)


Cadair:
As mentioned in chat we should have a way of parallelising over independent celestial input chunks as well.

astrofrog:
@Cadair - I am working on that but will add it in a separate PR. In the mean time, would you be able to review this or are you happy for me to just merge?

PR #376, opened 2023-07-14
Implement parallelization across broadcasted dimensions

astrofrog:
This builds on https://github.com/astropy/reproject/pull/374 to add a possible mode where if we are dealing with arrays with extra/broadcasted dimensions, we can split up the input array into chunks along the broadcasted dimensions which means we don't have to use the workaround of writing the whole dask array to a memmap.

For example, if we are reprojecting an array with dimensions (100, 2048, 2048) to (100, 1024, 1024), reprojecting only the celestial part of the WCS (the second and third dimensions of the array), we can specify e.g. ``block_size=(5, 1024, 1024)`` and this will cause the input to also be split up into chunks of ``(5, 2048, 2048)`` and each chunk of length 5 along the extra dimension will be handled separately.

This is very experimental at the moment and I need to do more experimenting to understand the performance - this doesn't always make things faster as the broadcasting is quite efficient normally whereas here the WCS transforms have to be repeated for each chunk, but on the plus side it means that for very large input arrays we never need to dump it to disk or compute it in one go.

Not really ready for testing yet!


astrofrog:
This works nicely now, obviously need to add tests etc but first going to see if I can fix co-add for arrays with broadcasting

PR #380, opened 2023-08-17
Add modes for nan and inf handling to adaptive algo

svank:
I found myself reprojecting images with NaNs and infs, so I added a few modes for how to deal with them. The default remains the previous behavior, which just propagates these values with no special handling.

The `constant` mode isn't strictly necessary, as one could use `np.nan_to_num` on the input array before reprojecting, but implementing this in `reproject` avoids having to copy the array---maybe that's important for `dask` use cases.

### DIFF
--- a/docs/celestial.rst
+++ b/docs/celestial.rst
@@ -288,7 +288,7 @@ will wrap around to the opposite side of the image, ignoring the
 ``boundary_mode`` for that axis.
 
 This implementation includes several options for handling ``nan`` and ``inf``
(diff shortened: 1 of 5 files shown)


astrofrog:
I've rebased this and have also changed ``bad_val_mode`` to ``bad_value_mode`` - I try and avoid abbreviations if it only saves a couple of letters. Hope that's ok! Will merge once CI passes.

svank:
@astrofrog Thanks for tying off these PRs! Happy to see another release :tada:

PR #382, opened 2023-08-26
Use more points to find image bounds in moasics

svank:
I'm trying out `reproject_and_coadd` (thanks @Cadair!). I'm making composite images from two WISPR images, which is together a pretty wide FOV (~100 degrees across). One of the two images is convex in the output projection, so checking where the image's corners fall in the output image isn't enough to properly bound it.

I've changed the function to put ten points along each edge of the image and use those to find its bounds in the output image, rather than just checking the corners. For my use-case, ten points per edge works just as well as doing every edge pixel.

Looking at just the footprint of the reprojected images, it's the difference between this (note the left edge):
image
and this:
image

### DIFF
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -134,7 +134,7 @@ def reproject_and_coadd(
         # convex in the output projection), and transforming every edge pixel,
         # which provides a lot of redundant information.
         ny, nx = array_in.shape
-        n_per_edge = 10
+        n_per_edge = 11
         xs = np.linspace(-0.5, nx - 0.5, n_per_edge)
         ys = np.linspace(-0.5, ny - 0.5, n_per_edge)
         xs = np.concatenate((xs, np.full(n_per_edge, xs[-1]), xs, np.full(n_per_edge, xs[0])))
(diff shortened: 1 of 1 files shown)


astrofrog:
What about making this value configurable via a kwarg? 10 seems a sensible default though.

svank:
Then I would have to make a test. :grimacing: But are you aware of projections that would require a lot more points? I'm not sure I've seen any that get truly wild---the smooth convexness in these WISPR images is the worst I can think of, and it seems well-served by a modest number of points along the edge. And I'm not sure there's a use case for turning the number down---whether we're doing the four corners or 40 points along the edges, the computational cost here seems truly negligible compared to the actual reprojection. So my thinking was to avoid a proliferation of options since I can't think of a good use case. But I can add the option if you'd rather.

Come to think of it, 9 or 11 might be the better default, so there's a point at the middle of each edge---it seems easy to imagine that being the most extreme point in cases where there's not also rotation at play, so might as well make sure to include it.

astrofrog:
@svank - sounds good about avoiding a kwarg. I'll push a change to make this 11 and will merge.

PR #383, opened 2023-08-26
Add 'first' and 'last' moasicking modes to reproject_and_coadd

svank:
Again with WISPR and `reproject_and_coadd`, merging the image data in the overlap region isn't the most desirable thing to do, for two reasons. First, the WISPR L3 images don't always have a well-matched background level (because of how much processing it takes to get to L3), so at the small overlap region between the two FOVs, taking a mean of the two images gives a visible seam on either side. Second, there's sometimes a time offset between images from the two images, so if I'm building the composite in HPC coordinates, the stars are offset by a few pixels and produce ghosting in the overlap region. (If I build the composite in RA/Dec coordinates, that should stop being a problem, but then I need to think about how that offsets the coronal structures and how to make an HPC WCS for my RA/Dec composite frame.)

The `match_background` option helps hide the seams, but it's not perfect.

This PR adds `first` and `last` options for `combine_function`, which provide a simple stacking mode. In overlap regions, either the first or the last of the input images that cover any pixel determines the output value for that pixel. For the WISPR case, this artistic choice makes one seam more obvious, but hides the second seam and the star misalignment.

With `combine_function="mean"` (make it full-size to see the star ghosting):
image

With `combine_function="first"`:
image

### DIFF
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -73,9 +73,12 @@ def reproject_and_coadd(
         `~astropy.io.fits.HDUList` instance, specifies the HDU to use.
     reproject_function : callable
         The function to use for the reprojection
-    combine_function : { 'mean', 'sum', 'median' }
+    combine_function : { 'mean', 'sum', 'median', 'first', 'last' }
         The type of function to use for combining the values into the final
-        image.
+        image. For 'first' and 'last', respectively, the reprojected images are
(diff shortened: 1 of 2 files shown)


astrofrog (approved):
Looks great, thanks!

astrofrog:
The wheel failures are unrelated, they are in Python 3.12 and not due to reproject itself.

PR #386, opened 2023-09-08
Fix TestReprojectAndCoAdd failure on i386

olebole:
On i386, this test [resulted in a failure](https://salsa.debian.org/debian-astro-team/reproject/-/jobs/4673706) because of relative differences ~2e-16. As floats are compared, minimal floating point differences my appear. This PR replaces the strict equal with an allclose test.


astrofrog:
pre-commit.ci autofix

PR #390, opened 2023-09-12
Fixed a bug that caused reprojected dask arrays to not be computable due to a temporary directory being removed

astrofrog:
@keflavich - this is a fix for the issue we were seeing. Note that we need to keep the temporary directory around for a while because the user might call compute multiple times etc. So I think we might just need to document how to clean up the temporary directory once it is no longer needed.

### DIFF
--- a/reproject/common.py
+++ b/reproject/common.py
@@ -100,7 +100,7 @@ def _reproject_dispatcher(
     # We set up a global temporary directory since this will be used e.g. to
     # store memory mapped Numpy arrays and zarr arrays.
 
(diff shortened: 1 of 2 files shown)


keflavich:
Yes, that will certainly be important. Will it naturally get cleaned up if the `cube` is deleted?  It's not obvious to me that it will.

astrofrog:
@keflavich - this won't get cleaned up automatically except by the operating system using the regular temporary directory cleanup. So this is not a good solution for spectral-cube. However, it's a fix that at least allows smaller cubes to be reprojected, so I'll go ahead and merge and work on a better solution.

PR #399, opened 2023-10-11
Add testing on Python 3.12

dstansby:
I can get this building locally fine on Python 3.12 (using `python -m build`), but can reproduce the tox failure. so I think the issue must be somehow `tox` is installing a different and older version of `Cython` when trying to build `reproject`. I'm a bit lost on how to debug that though...

Update: if I remove `aiohttp` from the deps (that is also trying and failing to build a wheel due to Cython issues) this works fine. So I guess there's some dogy Cython version interaction going on somewhere... I think the answer here is to wait for `aiohttp` to get a new version out with 3.12 support and hopefully this will Just Work ™️ after that.

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -18,16 +18,19 @@ jobs:
         - macos: py39-test
         - macos: py310-test
         - macos: py311-test
+        - macos: py312-test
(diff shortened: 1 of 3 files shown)


astrofrog:
Sorry I accidentally pushed these commits to ``main`` 🤦‍♂️  - I'll open a new PR

PR #403, opened 2023-10-19
Python 3.12 testing

dstansby:


dstansby:
Ha, so it turns out the issue isn't building the local copy of reproject, but the testing dep `sunpy[map]` pulling in a dependency on `reproject`, and trying to download and build the latest released verision from PyPI, which doesn't compile under Python 3.12...

dstansby:
So I think the best way round this is to not test with sunpy on Python 3.12, until a release of reproject has been done with Python 3.12 to PyPI. I can't see another good way to avoid the sunpy test dependency trying to pull in a version of reproject from PyPI that is incompatible with Python 3.12.

astrofrog (approved):
Looks good, thanks!

PR #418, opened 2023-11-28
Fix Cython warnings

### LINKED ISSUE #416: Cython performance warnings in deforest.pyx
@svank - now that we've updated the version of Cython used there are some performance hints:

```
performance hint: reproject/adaptive/deforest.pyx:54:5: Exception check on 'svd2x2_decompose' will always require the GIL to be acquired.
      Possible solutions:
          1. Declare the function as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
          2. Use an 'int' return type on the function to allow an error code to be returned.
      performance hint: reproject/adaptive/deforest.pyx:81:5: Exception check on 'mul2x2' will always require the GIL to be acquired.
      Possible solutions:
          1. Declare the function as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
          2. Use an 'int' return type on the function to allow an error code to be returned.
      performance hint: reproject/adaptive/deforest.pyx:100:5: Exception check on 'svd2x2_compose' will always require the GIL to be acquired.
      Possible solutions:
          1. Declare the function as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
          2. Use an 'int' return type on the function to allow an error code to be returned.
      performance hint: reproject/adaptive/deforest.pyx:189:5: Exception check on 'calculate_jacobian' will always require the GIL to be acquired.
      Possible solutions:
          1. Declare the function as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
          2. Use an 'int' return type on the function to allow an error code to be returned.
      performance hint: reproject/adaptive/deforest.pyx:230:5: Exception check on 'sort' will always require the GIL to be acquired.
      Possible solutions:
          1. Declare the function as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
          2. Use an 'int' return type on the function to allow an error code to be returned.
      performance hint: reproject/adaptive/deforest.pyx:320:5: Exception check on 'fill_in_jacobian' will always require the GIL to be acquired.
      Possible solutions:
          1. Declare the function as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.
          2. Use an 'int' return type on the function to allow an error code to be returned.
```

Do you think we should do something about these?

svank:
Fixes  #416

Implements Cython's suggested performance fixes and removes the "performance hint" warnings. Looks like [new behavior in Cython 3.0](https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#exception-values-and-noexcept), to ensure proper handling if a Python exception is raised within a function. Since these subroutines don't call any Python code directly, no Python exception can be raised, so we can add `noexcept` flags and avoid Cython checking for an exception after every function call. These changes gain a bit of speed in a real-world benchmark, but it looks like just speed that we lost when changing to Cython 3.

I also fixed a few warnings (removed an unused function, compiler worried about an uninitialized variable, [explicitly set `language_level` to the new default](https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#python-3-syntax-semantics), and [added flag to not use the deprecated numpy api](https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#numpy-c-api),) so now `deforest.pyx` builds without any warnings on my computer.

Performance of the adaptive code is indeed relevant to my interests---thanks for pointing this out, @astrofrog

### DIFF
--- a/reproject/adaptive/deforest.pyx
+++ b/reproject/adaptive/deforest.pyx
@@ -1,4 +1,5 @@
-#cython: boundscheck=False, wraparound=False, nonecheck=False, cdivision=True
+#cython: boundscheck=False, wraparound=False, nonecheck=False, cdivision=True, language_level=3str
+#distutils: define_macros=NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION
 
 # Cython implementation of the image resampling method described in "On
 # resampling of Solar Images", C.E. DeForest, Solar Physics 2004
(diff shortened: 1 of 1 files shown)


astrofrog (approved):

astrofrog:
Thanks!

PR #427, opened 2024-02-12
Fix 426: add blank_pixel_value keyword

keflavich:
This adds a `blank_pixel_value` keyword that defaults to `nan` and applies to all combination modes in the coadd/mosaicking step.


astrofrog:
This was done as part of https://github.com/astropy/reproject/pull/351 in the end

PR #439, opened 2024-04-15
Add a script to regenerate the aia asdf file and do so

Cadair:


Cadair:
It's not clear to me why the data would appear to have changed, but since I can't regenerate the original file I don't know how to see why!

astrofrog:
The data hasn't changed though, just the metadata, right?

astrofrog:
For the test failure related to the adaptive roundtrip test it looks like the sun rotates a bit between the new result and the reference result

[Screencast from 16-07-24 11:10:43.webm](https://github.com/user-attachments/assets/150ee3c3-4090-40f0-a96c-581c7501be2f)

astrofrog:
@Cadair - hopefully this should work now! Let's see what CI says.

PR #443, opened 2024-06-04
Improve performance for large datasets and switch to multi-threading by default

### LINKED ISSUE #393: Avoid writing out a temporary memmap if the input data can already be recognized as a memmap
Currently when in parallel mode, if a user specifies a filename for the input file we load the data with astropy.io.fits, which even if memory mapped is then written out to a new memory-mapped file for the purposes of the parallel computation (to avoid copying the array in memory to all the processes).

We should find a way, whenever possible, to avoid writing out a new memmap if the original data is backed by a file on disk.

I'm not sure if there is a way to do this if HDU objects are passed in as ``hdu.data`` is a regular Numpy array and the details of the memmap are hidden in the buffer. However if a filename is passed, we should be able to set up a memmap ourselves using the BITPIX and NAXIS? in the header for the HDU. If parallel mode is specified, we could then warn if an HDU or HDUList is passed that this is not optimal and that a filename should be passed instead.

We should also make sure we support passing in ``np.memmap`` objects and properly handle these (again avoiding any re-writing out of the arrays).

### LINKED ISSUE #394: Use multi-threading instead of multi-processing
The main functions in reproject, such as ``reproject_interp``, accept ``parallel=`` and ``block_size=`` arguments which, if used, will leverage dask behind the scenes to split up the data into chunks and then use a multi-**processing** scheduler to distribute the work.

Ideally we should be using multi-threading instead of multi-processing, but currently we don't because there appear to be some issues with some output pixels not having the right value when using multi-threading.

Fixing this will provide two main benefits:

* Avoid the whole dumping arrays to memmaps and being able to just use input arrays as-is without duplication
* Avoiding nasty surprises for users who select ``return_type='dask'`` then use the default scheduler to compute the array

Here's a notebook illustrating the issues: https://gist.github.com/astrofrog/e8808ee3ee8b7b86a979e0cb305d518b - note that while there is no explicit mention of threads anywhere, in the ``compare`` function ``array2`` is a dask array and when it gets passed to Matplotlib, ``.compute()`` gets called and the default scheduler uses threads.

At the moment all algorithms seem to have issues though all appear to be different. Note that for adaptive I sometimes have to run a few times to see issues.

astrofrog:
This is a work in progress to address serious performance issues when reprojecting large datasets, mainly with ``reproject_interp``.

To do:

* [x] Add more robust checks for ``hdu_to_vanilla_memmap`` to make sure we handle the case where it might fail (e.g. for an in-memory HDU)
* [x] Document TMPDIR and ``intermediate_memmap``
* [ ] Write up recommendations for if someone wants to run using dask distributed with multi-processing

Fixes #393 
Fixes #394

### DIFF
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -1,7 +1,7 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import os
+import sys
 import tempfile
-import time
 import uuid
 
 import numpy as np
@@ -15,6 +15,9 @@ from .background import determine_offset_matrix, solve_corrections_sgd
 from .subset_array import ReprojectedArraySubset
 
 __all__ = ["reproject_and_coadd"]
+
+
+IS_WIN = sys.platform == "win32"
 
 
 def _noop(iterable):
@@ -37,6 +40,7 @@ def reproject_and_coadd(
     block_sizes=None,
     progress_bar=None,
     blank_pixel_value=0,
+    intermediate_memmap=False,
     **kwargs,
 ):
     """
@@ -118,6 +122,9 @@ def reproject_and_coadd(
     blank_pixel_value : float, optional
         Value to use for areas of the resulting mosaic that do not have input
         data.
+    intermediate_memmap : bool, optional
+        If `True`, use `numpy.memmap` to store intermediate output arrays for
+        reprojected data.
 
     **kwargs
         Keyword arguments to be passed to the reprojection function.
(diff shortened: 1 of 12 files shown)


astrofrog:
Running into memory issues when mosaicking large datasets after the first one is done which suggests maybe an issue with the data getting loaded at the point where it is added to the main array.

reproject/mosaicking/coadd.py
```diff
@@ -245,23 +250,51 @@ def reproject_and_coadd(
...
         # footprint + weight map
 
-        array, footprint = reproject_function(
-            (array_in, wcs_in),
-            output_projection=wcs_out_indiv,
-            shape_out=shape_out_indiv,
-            hdu_in=hdu_in,
-            **kwargs,
-        )
+        with tempfile.TemporaryDirectory() as local_tmp_dir:
```
astrofrog, line 253: I think this needs to be around the whole loop probably

astrofrog:
Some CI failures related to zarr 3.0.0a1 released yesterday, need to investigate.

astrofrog:
I'll go ahead and merge this and will then check the multi-processing and document any steps needed in a separate PR.

PR #457, opened 2024-07-23
Fix CI following Sunpy 6.0.0 release

astrofrog:
With Sunpy 6.0.0, a couple of date-related keyword arguments are added to the WCS/header, we remove these here since they are not critical to the comparison with the reference files.


Cadair (approved):

astrofrog:
devdeps is broken because of a zarr/dask incompatibility, will need to investigate later but not critical

PR #459, opened 2024-07-25
Fix bug with artifacts in reproject_to_healpix

astrofrog:
These were due to using scipy's map_coordinates directly instead of our wrapper which deals properly with boundaries


Wasim04:
Thanks a lot for your support. I shall try to implement this. Many Thanks

Issue #471, opened 2024-09-27
BUG: CI is failing badly


pllim:
A lot of jobs failing.

astrofrog:
Last successful build of tests: https://github.com/astropy/reproject/actions/runs/10398602265

First failing one: https://github.com/astropy/reproject/actions/runs/10414456195

astrofrog:
Changes in dependencies:

```
2c2
< aiohappyeyeballs==2.3.5
---
> aiohappyeyeballs==2.3.6
29c29
< hypothesis==6.111.0
---
> hypothesis==6.111.1
60c60
< reproject @ file:///Users/runner/work/reproject/reproject/.tox/.tmp/package/1/reproject-0.14.1.dev8%2Bg7aaeb13a.tar.gz#sha256=c8bfb16dae88027ac4e0781f24d2bf9f84f90598892eea10d659f9d6b69caa32
---
> reproject @ file:///Users/runner/work/reproject/reproject/.tox/.tmp/package/1/reproject-0.14.1.dev8%2Bg7aaeb13a.tar.gz#sha256=fab9754e6ff7eaff955ad4de4ad2f4c97d806b56fb34793b61fa852f836c88d4
67c67
< sunpy==6.0.0
---
> sunpy==6.0.1
```

I wonder if this is related to sunpy

astrofrog:
Looks like it's due to sunpy 6.0.1: https://github.com/astropy/reproject/pull/472

astrofrog:
@Cadair could this be due to sunpy actually making something more correct so that we have to update the reference files?

PR #472, opened 2024-09-27
Fix Continuous Integration folllowing changes in Sunpy 6.0.0 and 6.0.1

astrofrog:
Experimental for now, aim is to solve https://github.com/astropy/reproject/issues/471


pyproject.toml
Cadair, line 1: should this also come with a test dep version bump for sunpy?

astrofrog: Good catch - fixed!

PR #474, opened 2024-10-01
Flag out NaN weights in coadd & allow for weights to have different WCS than data

keflavich:
I encountered a severe error case in which the weights array became NaN after reprojection.  Flagging out locations where weights is NaN solved the issue.

I cannot come up with an MWE; this occurs deep in the guts of a complicated mosaicking attempt, and it happened for only 1 field out of 270.  I think it might be the consequence of a numerical error that occurs only under unique conditions, but it also looks logically like this fix _should_ work.

### DIFF
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -25,6 +25,34 @@ def _noop(iterable):
     return iterable
 
 
+def byteordermatch(arr1, arr2):
+    """
+    Match arr1's dtype to arr2
+    """
+
+    odt = arr2.dtype
+    idt = arr1.dtype
+    bo_resolver = {'=': '<' if sys.byteorder == 'little' else '>', '<': '<', '>': '>'}
+    obo = bo_resolver[odt.byteorder]
+    ibo = bo_resolver[idt.byteorder]
+
+    if ibo != obo:
+        if hasattr(arr1, 'byteswap'):
+            arr1 = arr1.byteswap(inplace=True).view(arr2.dtype)
+        else:
+            from dask.utils import M
+
+            #arr1 = arr1.map_blocks(np.ndarray.byteswap, True).map_blocks(np.ndarray.newbyteorder, "S")
+            try:
+                arr1 = arr1.map_blocks(M.byteswap, False).map_blocks(M.newbyteorder, "S")
+            except:
+                def newbyteorder(arr, order):
+                    return arr.view(arr.dtype.newbyteorder(order))
+                arr1 = arr1.map_blocks(M.byteswap, False).map_blocks(newbyteorder, "S")
+
+    return arr1
+
+
 def reproject_and_coadd(
     input_data,
     output_projection,
@@ -171,6 +199,7 @@ def reproject_and_coadd(
             f"the output shape {shape_out}"
         )
 
+
     if output_footprint is None:
         output_footprint = np.zeros(shape_out)
     elif output_footprint.shape != shape_out:
@@ -206,6 +235,7 @@ def reproject_and_coadd(
             # We need to pre-parse the data here since we need to figure out how to
             # optimize/minimize the size of each output tile (see below).
             array_in, wcs_in = parse_input_data(input_data[idata], hdu_in=hdu_in)
+            array_in = byteordermatch(array_in, output_array)
 
             # We also get the weights map, if specified
             if input_weights is None:
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -25,34 +25,6 @@ def _noop(iterable):
     return iterable
 
 
-def _byteordermatch(arr1, arr2):
-
-    # Match arr1's dtype to arr2
-
-    odt = arr2.dtype
-    idt = arr1.dtype
-    bo_resolver = {"=": "<" if sys.byteorder == "little" else ">", "<": "<", ">": ">"}
-    obo = bo_resolver[odt.byteorder]
-    ibo = bo_resolver[idt.byteorder]
-
-    if ibo != obo:
-        if hasattr(arr1, "byteswap"):
-            arr1 = arr1.byteswap(inplace=True).view(arr2.dtype)
-        else:
-            from dask.utils import M
-
-            try:
-                arr1 = arr1.map_blocks(M.byteswap, False).map_blocks(M.newbyteorder, "S")
-            except Exception:
-
-                def newbyteorder(arr, order):
-                    return arr.view(arr.dtype.newbyteorder(order))
-
-                arr1 = arr1.map_blocks(M.byteswap, False).map_blocks(newbyteorder, "S")
-
-    return arr1
-
-
 def reproject_and_coadd(
     input_data,
     output_projection,
@@ -234,7 +206,6 @@ def reproject_and_coadd(
             # We need to pre-parse the data here since we need to figure out how to
             # optimize/minimize the size of each output tile (see below).
             array_in, wcs_in = parse_input_data(input_data[idata], hdu_in=hdu_in)
-            array_in = _byteordermatch(array_in, output_array)
 
             # We also get the weights map, if specified
             if input_weights is None:
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -358,7 +358,7 @@ def reproject_and_coadd(
 
             # For the purposes of mosaicking, we mask out NaN values from the array
             # and set the footprint to 0 at these locations.
-            reset = np.isnan(array)
+            reset = np.isnan(array) | np.isnan(weights)
             array[reset] = 0.0
             footprint[reset] = 0.0
 
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -212,7 +212,7 @@ def reproject_and_coadd(
                 weights_in = None
             else:
                 weights_in, weights_wcs = parse_input_weights(input_weights[idata], hdu_weights=hdu_weights, return_wcs=True)
-                if weights_wcs is None:
+                if weights_wcs is None or not weights_wcs.has_celestial:
                     # if weights are passed as an array
                     weights_wcs = wcs_in
                 if np.any(np.isnan(weights_in)):
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -32,22 +32,24 @@ def byteordermatch(arr1, arr2):
 
     odt = arr2.dtype
     idt = arr1.dtype
-    bo_resolver = {'=': '<' if sys.byteorder == 'little' else '>', '<': '<', '>': '>'}
+    bo_resolver = {"=": "<" if sys.byteorder == "little" else ">", "<": "<", ">": ">"}
     obo = bo_resolver[odt.byteorder]
     ibo = bo_resolver[idt.byteorder]
 
     if ibo != obo:
-        if hasattr(arr1, 'byteswap'):
+        if hasattr(arr1, "byteswap"):
             arr1 = arr1.byteswap(inplace=True).view(arr2.dtype)
         else:
             from dask.utils import M
 
-            #arr1 = arr1.map_blocks(np.ndarray.byteswap, True).map_blocks(np.ndarray.newbyteorder, "S")
+            # arr1 = arr1.map_blocks(np.ndarray.byteswap, True).map_blocks(np.ndarray.newbyteorder, "S")
             try:
                 arr1 = arr1.map_blocks(M.byteswap, False).map_blocks(M.newbyteorder, "S")
             except:
+
                 def newbyteorder(arr, order):
                     return arr.view(arr.dtype.newbyteorder(order))
+
                 arr1 = arr1.map_blocks(M.byteswap, False).map_blocks(newbyteorder, "S")
 
     return arr1
@@ -199,7 +201,6 @@ def reproject_and_coadd(
             f"the output shape {shape_out}"
         )
 
-
     if output_footprint is None:
         output_footprint = np.zeros(shape_out)
     elif output_footprint.shape != shape_out:
@@ -241,7 +242,9 @@ def reproject_and_coadd(
             if input_weights is None:
                 weights_in = None
             else:
-                weights_in, weights_wcs = parse_input_weights(input_weights[idata], hdu_weights=hdu_weights, return_wcs=True)
+                weights_in, weights_wcs = parse_input_weights(
+                    input_weights[idata], hdu_weights=hdu_weights, return_wcs=True
+                )
                 if weights_wcs is None or not weights_wcs.has_celestial:
                     # if weights are passed as an array
                     weights_wcs = wcs_in
--- a/reproject/mosaicking/coadd.py
+++ b/reproject/mosaicking/coadd.py
@@ -25,10 +25,9 @@ def _noop(iterable):
     return iterable
 
 
-def byteordermatch(arr1, arr2):
-    """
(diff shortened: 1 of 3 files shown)


keflavich:
Currently broken:
```
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[interp-False-arrays] - TypeError: input_data should either be an HDU object or a tuple of (array, WCS) or (array, Header)
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[interp-False-filenames] - ValueError: Output WCS has celestial components but input WCS does not
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[interp-False-hdus] - ValueError: Output WCS has celestial components but input WCS does not
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[interp-True-arrays] - TypeError: input_data should either be an HDU object or a tuple of (array, WCS) or (array, Header)
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[interp-True-filenames] - ValueError: Output WCS has celestial components but input WCS does not
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[interp-True-hdus] - ValueError: Output WCS has celestial components but input WCS does not
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[exact-False-arrays] - TypeError: input_data should either be an HDU object or a tuple of (array, WCS) or (array, Header)
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[exact-False-filenames] - NotImplementedError: Currently only data with a 2-d celestial WCS can be reprojected using flux-conserving algorithm
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[exact-False-hdus] - NotImplementedError: Currently only data with a 2-d celestial WCS can be reprojected using flux-conserving algorithm
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[exact-True-arrays] - TypeError: input_data should either be an HDU object or a tuple of (array, WCS) or (array, Header)
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[exact-True-filenames] - NotImplementedError: Currently only data with a 2-d celestial WCS can be reprojected using flux-conserving algorithm
FAILED reproject/mosaicking/tests/test_coadd.py::TestReprojectAndCoAdd::test_coadd_with_weights[exact-True-hdus] - NotImplementedError: Currently only data with a 2-d celestial WCS can be reprojected using flux-conserving algorithm
FAILED reproject/mosaicking/tests/test_coadd.py::test_coadd_solar_map - TypeError: input_data should either be an HDU object or a tuple of (array, WCS) or (array, Header)
```

keflavich:
Tests now pass locally (w/o sunpy installed...)

keflavich:
@astrofrog could you review this?  There are several critical fixes for spectral-cube's cube mosaicking.  

Major changes are:
 * Weights are no longer assumed to be on the same grid as the data, and they can bring their own WCS
 * input and output array dtypes are not assumed to be the same; input is swapped to match output (under the assumption that that's cheaper - but perhaps this needs to be made optional / documented)
 * bad weights are removed, which was the original point of this PR, but might have been a side-effect of the first

keflavich:
For coverage checks: Do we have a dask fits reader, or could we add that in?  The byte swapping does get tested in spectral-cube tests, since those sometimes load different endian arrays, but I don't see a simple way to do that here.

astrofrog (commented):
This looks mostly good to me, but one thing I'm not sure about is whether the byteswapping is happening too early here. Could you explain what issues you have been seeing here related to this? The byteswapping should probably ideally happen inside each individual reprojection chunk in case the input isn't a dask array and to avoid doubling memory usage?

keflavich:
> in case the input isn't a dask array

Technically I think this might be needed for numpy arrays too, but the FITS reader for numpy arrays already handles that?


In any case, I'll report the issues and see if we can move the byteswapping.  I haven't worked up a minimal example for the dask byteswapped data; iirc, it was any dask array loaded with a fits reader

keflavich:
it doesn't crash.  It interprets the data as having the wrong endianness and all values end up as 2e-205 and similar.

astrofrog:
Ok interesting, is that making use of reproject_interp?

keflavich:
yes

keflavich:
so, I didn't figure out how to build a MWE within reproject, but if you comment out the `_byteordermatch` call and run the spectral cube tests, you'll get errors like this:
```
spectral_cube/tests/test_regrid.py ...................................................F.FFF............F.F.F.F.......                                                                                                                   [100%]

================================================================================================================== FAILURES ===================================================================================================================
_____________________________________________________________________________________________________ test_mosaic_cubes[True-100-False0] ______________________________________________________________________________________________________

use_memmap = False, data_adv = PosixPath('/private/var/folders/k_/7qh4l0nn72b7qgq15pkd4hw40000gt/T/pytest-of-adam/pytest-171/test_mosaic_cubes_True_100_Fal0/adv.fits'), use_dask = True, spectral_block_size = 100

    @pytest.mark.parametrize('spectral_block_size,use_memmap', ((None, False),
                                                                (100, False),
                                                                (None, True),
                                                                (100, False),
                                                                (1, True),
                                                                (1, False),
                                                                ))
    def test_mosaic_cubes(use_memmap, data_adv, use_dask, spectral_block_size):

        pytest.importorskip('reproject')

        # Read in data to use
        cube, data = cube_and_raw(data_adv, use_dask=use_dask)

        # cube is doppler-optical by default, which uses the rest wavelength,
        # which isn't auto-computed, resulting in nan pixels in the WCS transform
        cube._wcs.wcs.restwav = constants.c.to(u.m/u.s).value / cube.wcs.wcs.restfrq

        expected_header = combine_headers(cube.header, cube.header)
        expected_wcs = WCS(expected_header).celestial

        # Make two overlapping cubes of the data
        part1 = cube[:, :round(cube.shape[1]*2./3.), :]
        part2 = cube[:, round(cube.shape[1]/3.):, :]

        assert part1.wcs.wcs.restwav != 0
        assert part2.wcs.wcs.restwav != 0

        # Mosaic give the expected header.
        result = mosaic_cubes([part1, part2],
                              # order is not a mosaic_cubes argument order='nearest-neighbor',
                              target_header=cube.header,
                              # not used roundtrip_coords=False,
                              spectral_block_size=spectral_block_size,
                              save_to_tmp_dir=False,
                              verbose=True,
                              use_memmap=use_memmap,
                              method='cube')

        # Check that the shapes are the same
        assert result.shape == cube.shape

        assert repr(cube.wcs.celestial) == repr(result.wcs.celestial)

        # Check WCS in reprojected matches wcs_out
        # (comparing WCS failed for no reason we could discern)
        assert repr(expected_wcs) == repr(result.wcs.celestial)

        # Check that values of original and result are comparable
>       np.testing.assert_almost_equal(result.filled_data[:].value,
                                       cube.filled_data[:].value, decimal=9)

spectral_cube/tests/test_regrid.py:651:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../mambaforge/envs/py312/lib/python3.12/contextlib.py:81: in inner
    return func(*args, **kwds)
../../mambaforge/envs/py312/lib/python3.12/contextlib.py:81: in inner
    return func(*args, **kwds)
../../mambaforge/envs/py312/lib/python3.12/site-packages/numpy/_utils/__init__.py:85: in wrapper
    return fun(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

args = (<function assert_array_almost_equal.<locals>.compare at 0x284f77920>, array([[[ 2.68308185e+185, -2.51563734e+252],
 ...0.39923534]],

       [[0.74684203, 0.85496775],
        [0.65468046, 0.96629004],
        [0.64365533, 0.74426717]]]))
kwds = {'err_msg': '', 'header': 'Arrays are not almost equal to 9 decimals', 'precision': 9, 'verbose': True}

    @wraps(func)
    def inner(*args, **kwds):
        with self._recreate_cm():
>           return func(*args, **kwds)
E           AssertionError:
E           Arrays are not almost equal to 9 decimals
E
E           Mismatched elements: 24 / 24 (100%)
E           Max absolute difference among violations: 3.89622506e+287
E           Max relative difference among violations: 4.55716028e+287
E            ACTUAL: array([[[ 2.683081851e+185, -2.515637336e+252],
E                   [-9.039787586e+039,  3.043705986e+222],
E                   [-1.223304808e+051,  2.728961237e+167]],...
E            DESIRED: array([[[0.214686889, 0.909502371],
E                   [0.817235373, 0.903880475],
E                   [0.70687927 , 0.118076201]],...

../../mambaforge/envs/py312/lib/python3.12/contextlib.py:81: AssertionError
------------------------------------------------------------------------------------------------------------ Captured stdout call -------------------------------------------------------------------------------------------------------------
INFO: Using memory [spectral_cube.cube_utils]
INFO: Using Cube method [spectral_cube.cube_utils]
-------------------------------------------------------------------------------------------------------------- Captured log call --------------------------------------------------------------------------------------------------------------
INFO     astropy:cube_utils.py:1004 Using memory
INFO     astropy:cube_utils.py:1004 Using Cube method
```

astrofrog:
Ok perfect, thanks!

astrofrog:
Ok I have a MWE with just ``reproject_interp``, so I'll see how we can fix it.

astrofrog:
Ok I've spotted the bug, I will open a separate PR shortly

astrofrog:
@keflavich - see https://github.com/astropy/reproject/pull/487

astrofrog:
You should be able to rebase against main now as I have merged the other PR

astrofrog:
(Once you rebase this I will review and merge ASAP)

keflavich:
rebase done

reproject/mosaicking/coadd.py
```diff
@@ -211,7 +211,12 @@ def reproject_and_coadd(
             if input_weights is None:
                 weights_in = None
             else:
-                weights_in = parse_input_weights(input_weights[idata], hdu_weights=hdu_weights)
+                weights_in, weights_wcs = parse_input_weights(
+                    input_weights[idata], hdu_weights=hdu_weights, return_wcs=True
+                )
+                if weights_wcs is None or not weights_wcs.has_celestial:
```
astrofrog, line 217: ``weights_wcs.has_celestial`` breaks support for APE 14 WCS

reproject/mosaicking/coadd.py
```diff
@@ -211,7 +211,12 @@ def reproject_and_coadd(
             if input_weights is None:
                 weights_in = None
             else:
-                weights_in = parse_input_weights(input_weights[idata], hdu_weights=hdu_weights)
+                weights_in, weights_wcs = parse_input_weights(
+                    input_weights[idata], hdu_weights=hdu_weights, return_wcs=True
```
astrofrog, line 215: I think this is the only place parse_input_weights is used so I don't think we need the kwarg, we can just always return the WCS

reproject/utils.py
```diff
@@ -208,11 +209,19 @@ def parse_input_weights(input_weights, hdu_weights=None):
                 )
             else:
                 hdu_weights = 0
-        return parse_input_data(input_weights[hdu_weights])[0]
+        return parse_input_data(input_weights[hdu_weights])[slc]
     elif isinstance(input_weights, PrimaryHDU | ImageHDU | CompImageHDU):
-        return input_weights.data
+        if return_wcs:
+            ww = WCS(input_weights.header)
+            ww = ww if ww.has_celestial else None
```
astrofrog, line 216: Here we again break compatibility with APE 14 WCS - why is it needed?

astrofrog (changes requested):
This looks mostly good but a couple of comments - in particular we shouldn't break APE-14 support. I don't think you need to check ``has_celestial``, if the weight's WCS isn't compatible with the mosaic WCS then an error will happen during reprojection anyway. We shouldn't ignore the WCS just because it's not celestial?

keflavich:
`has_celestial` isn't necessary, but we need something to check that there is a valid WCS.  The default is just `WCS()`, and it is possible (likely) for a user to pass a FITS HDU with a `weight` array and no WCS information in the header.  What's the appropriate way to do this check under APE14?

reproject/mosaicking/coadd.py
keflavich: fine by me, i can remove the kwarg

keflavich:
Your suggestion seems to be to just let it break downstream if there is an invalid WCS?  That's fine, I guess.  I'm not a fan of that strategy in general; I like to catch obvious errors where I know they'll occur.

astrofrog:
@keflavich - I think ``reproject_and_coadd`` can be used for any arbitrary WCS that might not even be celestial? Or is that incorrect?

keflavich:
right.... the cube modifications made that possible.  OK, I'll get rid of has_celestial, but I think the error messages may be opaque

keflavich:
Also, `ValueError: Output WCS has celestial components but input WCS does not`.  That's the expected error; I haven't looked more closely yet but at least the error is clear.

reproject/utils.py
keflavich: This check is serving a critical role and cannot be removed, so we have to think of a replacement.

This check is asking whether the FITS header contains a WCS that should be interpreted as the WCS for the weights.  Since `WCS()` (i.e., `WCS` with an empty dict as input) is still a valid WCS object but contains no transforms, we need some check here whether the WCS exists.    Annoyingly, you can't just check that `naxis` matches `ndim`, since `naxis` defaults to 2.

Maybe...
```python
ww = ww if ww.array_shape == input_weights.data.shape
```
...I'll check this...

keflavich: no, I don't think that works because the shape comes from the data when reading.

keflavich: what is the APE14 way to check if there is a transformation associated with an axis?

keflavich: I don't think there's any general way to get at this.  All WCSes contain valid transforms, so the only solution is to use the full WCS validation.  I've added that now.  It's less clear for specific cases, but at least it solves the general case.

astrofrog:
pre-commit.ci autofix

astrofrog (approved):

astrofrog:
@keflavich - sorry for the long delay in getting this in, this all looks good now!

PR #487, opened 2025-02-01
Fix a bug that caused big-endian dask arrays to not be reprojected correctly

astrofrog:
I'm surprised I didn't run into this sooner, but this should fix the endianness issue @keflavich - can you confirm?

### DIFF
--- a/reproject/utils.py
+++ b/reproject/utils.py
@@ -53,7 +53,7 @@ def _dask_to_numpy_memmap(dask_array, tmp_dir):
 
         memmapped_array = np.memmap(
             memmap_path,
-            dtype=zarr_array.dtype,
+            dtype=float,
             shape=zarr_array.shape,
             mode="w+",
         )
--- a/reproject/utils.py
+++ b/reproject/utils.py
@@ -31,6 +31,11 @@ def _dask_to_numpy_memmap(dask_array, tmp_dir):
     if isinstance(dask_array.ravel()[0].compute(), da.Array):
         dask_array = dask_array.compute()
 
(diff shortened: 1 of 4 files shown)


reproject/utils.py
```diff
@@ -53,7 +53,7 @@ def _dask_to_numpy_memmap(dask_array, tmp_dir):
 
         memmapped_array = np.memmap(
             memmap_path,
-            dtype=zarr_array.dtype,
+            dtype=float,
```
astrofrog, line 56: The memmap was read later on assuming it was native ``float``.

astrofrog:
I think I need to check the case where ``output_array`` is passed and it has a different itemsize to the input

astrofrog:
But for now I think it should work @keflavich

keflavich:
Yes, this solves the problem; I didn't realize there was a dask reader/converter built in to reproject.

I'll rebase #474 against this and remove the hack I implemented.

reproject/utils.py
```diff
@@ -31,6 +31,11 @@ def _dask_to_numpy_memmap(dask_array, tmp_dir):
     if isinstance(dask_array.ravel()[0].compute(), da.Array):
         dask_array = dask_array.compute()
 
+    # Cast the dask array to regular float for two reasons - first, zarr 3.0.0
+    # and later doesn't support big-endian arrays, and also we need to anyway
+    # make a native float memory mapped array below.
+    dask_array = dask_array.astype(float, copy=False)
```
olebole, line 37: Naive question: if zarr does not support big endian arrays, shouldn't this explicitly casted to a little endian float? On s390x, `float` is big endian.

Issue #491, opened 2025-03-11
Test failures on big endian machines


olebole:
This is a follow-up to #487, as it does not fix the issues with big endian machines. First, there are many failures that look like
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <enum 'DataType'>, dtype = dtype('float64')

    @classmethod
    def from_numpy(cls, dtype: np.dtype[Any]) -> DataType:
        if dtype.kind in "UT":
            return DataType.string
        elif dtype.kind == "S":
            return DataType.bytes
        elif not _NUMPY_SUPPORTS_VLEN_STRING and dtype.kind == "O":
            # numpy < 2.0 does not support vlen string dtype
            # so we fall back on object array of strings
            return DataType.string
        dtype_to_data_type = {
            "|b1": "bool",
            "bool": "bool",
            "|i1": "int8",
            "<i2": "int16",
            "<i4": "int32",
            "<i8": "int64",
            "|u1": "uint8",
            "<u2": "uint16",
            "<u4": "uint32",
            "<u8": "uint64",
            "<f2": "float16",
            "<f4": "float32",
            "<f8": "float64",
            "<c8": "complex64",
            "<c16": "complex128",
        }
>       return DataType[dtype_to_data_type[dtype.str]]
E       KeyError: '>f8'

/usr/lib/python3/dist-packages/zarr/core/metadata/v3.py:704: KeyError
```
See [this build log](https://buildd.debian.org/status/fetch.php?pkg=reproject&arch=s390x&ver=0.14.1-2&stamp=1741711339&file=log).
They are caused by trying to cast a native array into zarr, which is not supported for big endian. The patch

```
diff --git a/reproject/common.py b/reproject/common.py
index e4f2967..5f4cf5d 100644
--- a/reproject/common.py
+++ b/reproject/common.py
@@ -300,7 +300,7 @@ def _reproject_dispatcher(
             reproject_single_block,
             array_out_dask,
             array_in_or_path,
-            dtype=float,
+            dtype="<f8",
             new_axis=0,
             chunks=(2,) + array_out_dask.chunksize,
         )
diff --git a/reproject/utils.py b/reproject/utils.py
index b542885..6da8503 100644
--- a/reproject/utils.py
+++ b/reproject/utils.py
[...]

PR #492, opened 2025-03-11
Force little endian input for zarr

olebole:
I think I found the cause for the failures in #491: One is what was already mentioned in https://github.com/astropy/reproject/pull/487#pullrequestreview-2675588760, and the other is similar in `reproject/common.py`. The reason why the patch shown in #491 didn't work was that the second patch in `reproject/utils.py` was actually wrong.

I applied this PR to the Debian package (merged with #487), and now [all tests pass on s390x](https://buildd.debian.org/status/fetch.php?pkg=reproject&arch=s390x&ver=0.14.1-3&stamp=1741727132&raw=0).

Disclaimer: This is a fix by a "poor stranger". I didn't dig into the structure of reproject for this but just fixed what could cause the failures. There may be other places with a similar issue, or side effects that are not covered by the CI tests. So, please carefully review whether this does what is intended, beside the CI.

Closes: #491

### DIFF
--- a/reproject/common.py
+++ b/reproject/common.py
@@ -300,7 +300,7 @@ def _reproject_dispatcher(
             reproject_single_block,
             array_out_dask,
             array_in_or_path,
-            dtype=float,
+            dtype="<f8",
             new_axis=0,
             chunks=(2,) + array_out_dask.chunksize,
         )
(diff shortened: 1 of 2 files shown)


astrofrog:
Thank you for the PR! I am first going to try and add a big-endian CI job to try and catch this kind of issue in future.

PR #493, opened 2025-03-12
EXP: exotic architectures in CI

astrofrog:


astrofrog:
Well this was useless as it passes and doesn't reproduce https://github.com/astropy/reproject/issues/491

PR #498, opened 2025-04-28
Avoid using `astropy.utils.isiterable()`

eerovaher:
`astropy` will deprecate the function (in either 7.1 or 7.2), but it can be safely replaced with a `numpy` equivalent.


astrofrog (approved):
Thanks!

Issue #529, opened 2025-07-05
HIPS - how do you handle transparency?


keflavich:
I have tried HIPSing some png images with transparency, and they're coming out black.  Can we / how does HIPS support transparency?

keflavich:
I can't see where this is happening, but I'm continually reproducing it; alpha=0 is being transformed to alpha=255.   My best guess is that the problem lies in these lines:
https://github.com/astropy/reproject/blob/204bfd9ff01eea9be0e9d6745c7c4b42d28357c0/reproject/utils.py#L341-L345
but I can't see how the wrong logic would be adopted, i.e., how the input RGBA file would accidentally get the alpha layer dropped.

astrofrog:
@keflavich - is this still an issue if you upgrade pillow?

keflavich:
Yes, version 11.3 fixes it.  11.2 had the problem.

PR #530, opened 2025-07-05
Fix transparency preservation issue 529

keflavich:
Fixes Issue #529 

Just needed to force mode='RGBA'.

(also smuggled in a minor code correction)


astrofrog:
Thanks! Interesting, I removed mode because it is deprecated (see failing CI). I wonder how we are supposed to fix it now without mode.

keflavich:
This issues seems to be fixed with pillow==11.3

PR #539, opened 2025-07-22
Make it so we can use broadcasting even if input WCS is same dimension as data

astrofrog:
This is very much a WIP and not ready for usage but just opening this to not lose my progress.

I think we need to write down the rules for block sizes, and broadcasted reprojection, and I think it's ok to not cover all arbitrary cases. I'm now using the terminology 'reprojected dimensions' and 'non-reprojected dimensions', where the latter are essentially dimensions where we assume the mapping is 1-to-1 between input and output.

Proposed rules:

* The dimensionality of the overall reprojection is determined by the shape of the input data (the array itself, not the input data). Call this ``ndim``.
* If ``wcs_in`` has fewer pixel dimensions than ``ndim``, then we assume that the input WCS applies to the last ``wcs_in.pixel_n_dim`` dimensions of the input data, and that leading dimensions are the non-reprojected ones. The same applies to ``wcs_out``.
* ``shape_out`` should either have ``ndim`` elements, or match the number of reprojected dimensions. If the latter, then any missing elements for non-reprojected dimensions should be set to be the same as the input.
* ``block_size``, if specified, should also either have ``ndim`` elements, or match the number of reprojected dimensions. If the latter, then any missing elements for non-reprojected dimensions should be either set to ``1`` if ``block_size`` equals ``shape_out`` for reprojected dimensions, or ``-1`` otherwise
* ``block_size`` should either match ``shape_out`` for reprojected or for non-reprojected dimensions (that is, we either chunk over non-reprojected dimensions only, or over reprojected dimensions only). If it matches ``shape_out`` for reprojected dimensions, then the remaining block_size should be 1 so that only a single slice is reprojected at a time (yes there may be cases where someone wants to process slices in e.g. time or spectral slices, but this is going to introduce more complexity when the input WCS has dimension ``ndim`` so we punt on this for now).
* If specified, ``non_reprojected_dims`` should for now be only leading dimensions starting from ``0``, so e.g. ``(0,)``, ``(0, 1)`` and so on. ``(1,2)`` is not allowed for now. This could be relaxed in future potentially at the cost of more complexity.
* In principle one could imagine examples where different dimensions need to be ignored in the input and output, for example if the WCS order is different. I don't really want to cross that bridge, but if we ever do want to do this, the ``non_reprojected_dims`` option could take a list of tuples where each tuple gives the ``(input_dim, output_dim)`` correspondence.
* If ``non_reprojected_dims`` is specified, then if ``wcs_in`` or ``wcs_out`` have fewer dimensions than ``ndim``, the difference in the number of dimensions should match the length of ``non_reprojected_dims``.

For cases where the third axis is completely decoupled from the spatial axes in cubes, ``non_reprojected_dims`` isn't strictly needed because the WCSes could be easily sliced down to 2D. However, we do need that option for cases where the input WCS is 3D in the sense that each spatial slice might be different (for example due to drift at different times) but where each time still corresponds cleanly to one 2D slice and there is no dependence of e.g. time position on spatial position.

Complex examples that should work:

* If a 3D dataset is reprojected from a 3D spectral WCS to a 3D time WCS, if the spectral and time axes are the third WCS axis (first numpy axis) then if ``non_reprojected_dims=(0,)``, it should work and basically only care about the spatial to spatial conversion. This example doesn't make a huge amount of sense, but another more realistic example is that if one wants to align two spectral cubes spatially without touching the spectral axis, ``non_reprojected_dims=(0,)`` could do this.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -13,7 +13,7 @@ This package has the following dependencies:
 
 * `Numpy <http://www.numpy.org/>`__ 1.20 or later
 
-* `Astropy <http://www.astropy.org/>`__ 5.0 or later
+* `Astropy <http://www.astropy.org/>`__ 6.0 or later
 
 * `Scipy <http://www.scipy.org/>`__ 1.5 or later
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ urls = {Homepage = "https://reproject.readthedocs.io"}
 requires-python = ">=3.11"
 dependencies = [
     "numpy>=1.23",
-    "astropy>=5.0",
+    "astropy>=6.0",
     "astropy-healpix>=1.0",
     "scipy>=1.9",
     "dask[array]>=2024.4.1",
--- a/reproject/_common.py
+++ b/reproject/_common.py
@@ -95,11 +95,15 @@ def _reproject_dispatcher(
         the block size automatically determined. If ``block_size`` is not
         specified or set to `None`, the reprojection will not be carried out in
         blocks.
-    non_reprojected_dims : tuple
-        Dimensions that should not be reprojected but instead for which a
-        1-to-1 mapping between input and output pixel space should be assumed.
-        By default, this is any leading extra dimensions if the input WCS has
-        fewer dimensions than the input data.
+    non_reprojected_dims : tuple, optional
+        Leading dimensions of the data that should not be reprojected but for
+        which a one-to-one mapping between input and output pixels is assumed,
+        given as a tuple of sequential integers starting from zero (e.g.
+        ``(0,)`` or ``(0, 1)``). If `None` (the default), any leading dimensions
+        for which the WCS has fewer dimensions than the data are treated this
+        way. Reprojecting fewer dimensions than the WCS currently requires a
+        ``block_size`` that matches the output shape along the reprojected
+        dimensions.
     array_out : `~numpy.ndarray`, optional
         An array in which to store the reprojected data.  This can be any numpy
         array including a memory map, which may be helpful when dealing with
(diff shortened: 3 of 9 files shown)


astrofrog:
Ok so we need to add proper support for non_reprojected_dims to ``reproject_and_coadd`` rather than the current hack. And then of course a lot of documentation!

reproject/mosaicking/coadd.py
```diff
@@ -257,7 +262,7 @@ def reproject_and_coadd(
             ndim_out = len(shape_out)
 
             # Determine how many extra broadcasted dimensions are present
-            n_broadcasted = len(shape_out) - wcs_in.low_level_wcs.pixel_n_dim
+            n_broadcasted = len(shape_out) - wcs_out.low_level_wcs.pixel_n_dim
```
astrofrog, line 265: Hack currently!

astrofrog:
Ok so this is now tidied up and should be functional and https://github.com/astropy/reproject/pull/611 contains the part of extending this to co-addition. I'll do another self-review and then mark this ready for review.

astrofrog:
@svank - tagged you for review in case you want to take a look, this is an extension of the work you started in https://github.com/astropy/reproject/pull/332 - here we can now specify explicitly which dimensions to ignore - for now only leading dimensions can be ignored, but I plan to generalize this. The idea here was to efficiently be able to reproject say a solar time+spatial cube where the spatial coordinates drift over time, so one can't actually use a simple 2D WCS for both input and output WCS - the slicing has to be different for each slice.

For now I've not added narrative docs because this work is split up over multiple PRs and I want to wait until this is in good shape and stress-tested before documenting it.

astrofrog:
I'm going to go ahead and merge this so I can continue work on https://github.com/astropy/reproject/pull/611, but @svank feel free to leave comments post-merge, I'll don't plan to release these changes soon.

svank:
I think the rules in the first post look good to me! If I have it right, there's no change to the simple case I was implementing, where I have 3 (ny, nx) images with identical WCSes, so I stack them into a (3, ny, nx) array and pass it in. This is adding careful thought about how this interacts with block sizes, and adding `non_reprojected_dims` for the more complex cases you describe.

Sounds valuable to me! I'll look at the code changes more next week.

Issue #523, opened 2025-07-04
Document ``return_type='pil_image'`` and add tests


astrofrog:
We need to properly document the ``return_type='pil_image'`` option and add tests for it.

PR #542, opened 2025-07-28
Clean up RGB code and API

astrofrog:
This removed the short-lived ``return_type='pil_image'`` (never released) because it's a bit ambiguous how we should handle the footprint, and it would make more sense for users to decide how they want to construct an RGB image with the output. We can still use RGB images with AVM as input though. I've also removed ``as_rgb_images`` which is not needed anymore.

Closes https://github.com/astropy/reproject/issues/523


astrofrog:
I think the CI failiure is a fluke and unrelated

PR #543, opened 2025-07-28
Allow dask arrays to not be written to memmap

astrofrog:
In cases where one uses parallelization over extra dimensions, this can be a lot faster and avoid large disk usage for the memmap


reproject/common.py
```diff
@@ -116,6 +117,18 @@ def _reproject_dispatcher(
...
+            * ``'memmap'``: write out the entire input dask array to a temporary
+              memory-mapped array. This requires enough disk space to store
+              the entire input array, but should avoid accidentally loading
+              the entire array into memory.
+            * ``'none'``: load the dask array into memory as needed. This may
+              result in the entire array being loaded into memory. However,
+              this can be efficient under two conditions: if the array easily
+              fits into memory (as this will then be faster than ``'memmap'``),
+              and when the data contains more dimensions than the input WCS and
+              the block_size is chosen to iterate over the extra dimensions.
```
Cadair, line 131: (Not in this PR necessarily) Would it possible to detect the latter case and do this automatically?

PR #532, opened 2025-07-10
WIP: add a way to do reprojection without dumping dask arrays to memory maps

astrofrog:
Very much just work in progress, not ready for testing

PR #554, opened 2025-09-26
Added 2-d HiPS dask array class and fix a number of bugs with HiPS generation

astrofrog:
This includes a round-tripping test starting from an image, going to HiPS, and the reprojecting back through the dask class, which unveiled a number of bugs!

(this is extracted and modified from https://github.com/astropy/reproject/pull/514)


astrofrog:
I am not going to add narrative docs about this yet because for it to be truly useful in reproject we need to address https://github.com/astropy/reproject/pull/532

PR #587, opened 2026-03-03
EXP: Switch to meson for building package

astrofrog:
This is just an experiment and we don't have to switch to meson but I was curious what would be required and what kind of issues pop up


astrofrog:
As this was just an experiment, I am going to close this (not a huge benefit from using meson here as build time doesn't get way shorter)

PR #588, opened 2026-03-07
Support 32-bit floats in reproject_adaptive

svank:
I'd like to get eyes on this. In the PUNCH pipeline, I'm experimenting with using 32-bit floats to cut our in-memory data volume, and with having `reproject` put the output bits directly into a slice of a shared-memory array, so I can reproject a stack of images in parallel without having to copy inputs and outputs through pickle/multiprocessing. The latter doesn't seem to need any changes in `reproject` unless I want that array to be 32-bits.

To add single-precision support to `reproject_adaptive`, Cython has a `floating` type, where it generates both single- and double-precision version of the function. All instances of `floating` must match, though, so we can pass input and output arrays that are both 32-bits or both 64-bits, but not a mixture.

Then it just takes some care to make sure those input and output arrays always have the same dtype, and that that type is acceptable to cython. (If the input array has non-native endianness, that has to be fixed, for example.)


reproject/adaptive/core.py
```diff
@@ -117,8 +117,15 @@ def _reproject_adaptive_2d(
...
 
-    # Make sure image is floating point
-    array_in = np.asarray(array, dtype=float)
+    # Make sure the input is floating point, and that it matches the output array type if specified. (Cython makes 32
+    # and 64-bit versions of the inner function, but it's either everything is 32-bit, or everything is 64-bit). In
+    # case of a disagreement between the input and output types, it's best to apply the output's type to the
+    # input---the input is more likely to have weirdness (e.g. the wrong endianness, which Cython can't handle), and
+    # the output type is either a safe default, if the user didn't provide an output array, or it has a user-specified
+    # type.
+    array_in = np.asarray(array)
```
astrofrog: Does this change have any impact on what happens if someone calls reproject with an integer array?

svank: I think we'll still be good. `array_out` is either the default of `double`, or it's a user-provided array. In the latter case, if its dtype is `int` or otherwise weird, we can't do anything---we can't change or replace the user-provided array, so we'll crash. If `array_out` has a floating-point dtype, then we cast `array_in` to match, so that will handle integer `array_in`. (The first `as_array` is to make sure it's an array at all, with a `.dtype`, and the second `as_array` casts it if necessary. I think I can simplify that though.)

svank: I just pushed a commit that avoids the two `np.asarray` calls and makes the casting a little more clear.

astrofrog:
pre-commit.ci autofix

astrofrog (approved):

Issue #199, opened 2019-10-30
reproject_exact() is not conserving flux


acikota:
Dear all,
 
I have an issue with the reproject_exact() function, which doesn't properly conserve the flux.

To demonstrate the issue, I prepared an example which you can download from my Dropbox:
https://www.dropbox.com/s/x1a2upsxft5iru0/example.zip?dl=1

I have a highly sampled image (input.fits), which contains 4 pixels of values 10, 20, 30 and 40. When I reproject the image onto a template (templates/template.fits), the sum of the reprojected pixels is smaller by a factor of ~700.

I thought that the issue is produced by the difference in the area between the input image and the reprojected output image, i.e. that the flux is conserved per area, however, the pixel area ratio between template and input image is only ~226.
Furthermore, there is also an inconsistency in the pixel values, e.g. the pixels with the values 40 and 30 are equally bright in the reprojected image.

I would very appreciate your help or advice! 
Thank You!
 
Best regards,
Aleksandar

astrofrog:
@acikota - thanks for the bug report! I spent a little while investigating this, and it looks like this is a precision issue due to the very small pixels. I did some simple tests and it looks like things start to not work properly below a few times 0.01" in pixel sizes:

![ratios](https://user-images.githubusercontent.com/314716/67883201-8edcfb80-fb3b-11e9-948e-8c93f3993960.png)

For the record this plot can be made with:

```
# Check how good the reprojection is as a function of resolution

import numpy as np
from astropy.wcs import WCS
from astropy.io import fits
from reproject import reproject_exact

resolutions = np.logspace(-7, 0, 1000)
ratios = []

for res in resolutions:

    wcs1 = WCS()
    wcs1.wcs.ctype = 'RA---TAN', 'DEC--TAN'
    wcs1.wcs.crval = 13, 20
    wcs1.wcs.crpix = 10., 10.
    wcs1.wcs.cdelt = res, res

    wcs2 = WCS()
    wcs2.wcs.ctype = 'RA---TAN', 'DEC--TAN'
    wcs2.wcs.crval = 13, 20
    wcs2.wcs.crpix = 3, 3
    wcs2.wcs.cdelt = 3 * res, 3 * res

    array = np.zeros((19, 19))
    array[9, 9] = 1

    result, _ = reproject_exact((array, wcs1), wcs2, shape_out=(5, 5))

    ratios.append(9 * np.nansum(result) / np.nansum(array))

import matplotlib.pyplot as plt

[...]

PR #589, opened 2026-03-21
Improve accuracy for high-resolution images in reproject_exact

astrofrog:
As described in #199, the reproject_exact function suffers from precision issues when reprojecting images with sub-arcsecond pixel scales, which is particularly problematic for instruments like JWST NIRCam (0.03" pixels). There are two reasons for this: 

* Girard's theorem computes spherical polygon areas as sum_of_angles - (N-2)*π, which loses precision catastrophically when the polygon is tiny and the sum is nearly equal to (N-2)*π.
* The spherical polygon intersection algorithm in the relies on a TOLERANCE constant (~9e-4 arcsec) that causes incorrect intersection results when pixel features are smaller than this threshold.

This PR fixes both issues by adding a planar approximation path in computeOverlap. When all vertices of both the input and output pixels are within ~3.4 arcmin of their centroid, the code projects them onto a local tangent plane and performs the entire overlap computation in 2D — using Sutherland-Hodgman polygon clipping and the shoelace formula instead of the spherical intersection and Girard's theorem. This avoids all sources of precision loss, since coordinate values in the tangent plane are proportional to the polygon size rather than being tiny perturbations on a unit sphere. The same planar approximation is also applied as a fallback within the Girard function itself, to handle the edge case of tiny sliver overlaps between large pixels. With these changes, flux conservation is accurate down to ~1e-5 arcsec, compared to the previous limit of ~0.07 arcsec.

Here's an updated plot of accuracy vs resolution based on this PR:

ratios

For comparison, the original plot looked like:

### DIFF
--- a/reproject/spherical_intersect/core.py
+++ b/reproject/spherical_intersect/core.py
@@ -25,16 +25,16 @@ def _reproject_celestial(
     if output_footprint is None:
         output_footprint = np.empty(shape_out)
 
-    # There are precision issues below certain resolutions due to the
-    # TOLERANCE constant in the overlap area C code. For more details, see:
-    # https://github.com/astropy/reproject/issues/199
-    area_threshold = (1e-3 / 3600) ** 2
+    # There may be precision issues at extremely small resolutions due to
+    # floating-point limits in WCS coordinate transformations. For more
+    # details, see: https://github.com/astropy/reproject/issues/199
+    area_threshold = (1e-6 / 3600) ** 2
     if (isinstance(wcs_in, WCS) and proj_plane_pixel_area(wcs_in) < area_threshold) or (
         isinstance(wcs_out, WCS) and proj_plane_pixel_area(wcs_out) < area_threshold
     ):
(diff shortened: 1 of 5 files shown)


astrofrog:
@acikota - your example from Dropbox now returns:

```
pixel area ratio between template and input image:  226.47954059928594

reprojecting...
sum initial:  100.0
sum reprojected:  0.4415410594776611
ratio initial/reprojected sum: 226.47950366903376
```

which seems ok?

astrofrog:
So the interesting thing is that the planar approximation is actually better all the way up to 20" pixels. The comparison plot above looks like this if I zoom in to the y axis before the changes:

ratios

and like this after the changes:

ratios_new

So this means that actually for typical images the plane  approximation is actually better as the numerical precision improvements are more important than the algorithmic differences.

astrofrog:
I've corrected the calculation of the comparison at large resolutions, and the curves are now - before:

ratios


after:

ratios

PR #591, opened 2026-03-24
Clean up private/public API

astrofrog:
This PR moves many of the submodules which should be private to include leading underscores, and adds backward compatibility imports with deprecation warnings.

GitHub is not doing a good job at showing the changes - things are much clearer if you look at each individual commit.

@Cadair @svank - could you check what you think about this and whether this impacts any of the code/packages you have already?


Cadair (approved):

svank (approved):
No objection here!

PR #600, opened 2026-05-27
Enable limited API builds only when building wheels for publishing

### LINKED ISSUE #596: Allowing building for freethreaded Python
It looks like `reproject` cannot currently be built for freethreaded Python because `reproject` insists on the Limited API:
https://github.com/astropy/reproject/blob/c3c56d85de5af01d1fd8ac4e5ec6360f9be76119/pyproject.toml#L207-L208

Can an automatic switch be implemented a la [what is done in `pyerfa`](https://github.com/liberfa/pyerfa/blob/1dc9b120aa69cbfd29a14613c2fce35985006b26/setup.py#L22-L30) to support freethreading?

ping @Cadair

astrofrog:
This should fix https://github.com/astropy/reproject/issues/596 and allow local free-threaded builds

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -51,6 +51,8 @@ jobs:
         - cp*-macosx_x86_64
         - cp*-macosx_arm64
         - cp*-win_amd64
+      env: |
+        EXTENSION_HELPERS_PY_LIMITED_API: 'cp311'
 
       # Developer wheels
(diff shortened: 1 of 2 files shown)


astrofrog:
Seems to be working:

_Found previously built wheel reproject-0.19.1.dev53+g231c2e1d1-cp311-abi3-macosx_11_0_arm64.whl, that's compatible with cp312-macosx_arm64. Skipping build step..._

Cadair (approved):

PR #601, opened 2026-05-27
Added free-threaded CI jobs

astrofrog:
Needs to be rebased once https://github.com/astropy/reproject/pull/600 is merged

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -27,7 +27,6 @@ jobs:
           runs-on: ubuntu-24.04-arm
         - linux: py313-test
         - linux: py314-test-devdeps
-        - linux: py313t-test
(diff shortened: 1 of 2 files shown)


Cadair:
We should add a job using https://pypi.org/project/pytest-run-parallel/ using on both t and not t as it can find a bunch of thread races even on gil builds

astrofrog:
Sounds good - btw the existing failure here should be fixed by https://github.com/astropy/astropy/pull/19819

astrofrog:
Ok this now runs faster, I made some of the expensive tests single-threaded, and a few key tests be more intensive. This will need at least:

* https://github.com/astropy/astropy/pull/20424
* https://github.com/astropy/astropy/pull/20423

Before it can pass/be merged.

PR #613, opened 2026-07-01
Allow blocks smaller than full shape along reprojected dimensions and fix calculation of co-addition footprint in broadcasted parallelization

astrofrog:
WIP


astrofrog:
pre-commit.ci autofix

astrofrog:
I have decided not to go ahead with this - it has some downsides, including that there are cases where a full copy of the input will be made (e.g. with multiprocessing) and it's not as crucial now that we've fixed a number of memory issues in other PRs.

PR #616, opened 2026-07-03
Assemble each output chunk of the dask coaddition from only the images that overlap it and reduce memory usage for big-endian data

astrofrog:
This combines the pieces per chunk with the same per-pixel logic as the numpy path, so no zero padding chunks are materialized and the graph size and resident data scale with the true overlap instead of the number of images times the number of chunks.

### DIFF
--- a/reproject/mosaicking/_coadd.py
+++ b/reproject/mosaicking/_coadd.py
@@ -12,7 +12,7 @@ import numpy as np
 from astropy.wcs import WCS
 from astropy.wcs.wcsapi import SlicedLowLevelWCS
 
-from .._array_utils import iterate_chunks, pad_dask_array_to_grid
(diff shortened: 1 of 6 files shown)


reproject/mosaicking/_coadd.py
```diff
@@ -230,28 +230,103 @@ def _input_cutout_iterator(
         )
 
 
+def _combine_tile_pieces(
+    *pieces, combine_function, blank_pixel_value, dests, trailing_shape, block_info=None
+):
```
astrofrog, line 248: Can we perhaps merge with ``_combine_array_into_output``? TBD

astrofrog: This has been done

Issue #534, opened 2025-07-11
Reduce memory usage of pixel_to_pixel


astrofrog:
This isn't really an issue that requires work in reproject but it might require work in several other packages. Basically pixel_to_pixel can be very memory-intensive under certain situations. Here's an example:

https://gist.github.com/astrofrog/a807da0c34b7d669289e3c14d43a099a

this is with a GWCS and regular astropy WCS object. The input arrays are 80Mb each (160Mb total) but the memory peaks around 1.7-1.8Gb which is 10x more than this. It might be there's not much we can do about it, but it'd be interesting to see if there are places where we memory could be used more efficiently. Obviously, the output arrays also need 80Mb each, but that's still a lot of overhead.

Cadair:
oh good, looks like the next modeling task is memory optimisation lol

PR #617, opened 2026-07-04
Significantly reduce memory usage due to WCS transformations

astrofrog:
This does several things to reduce memory usage:

* It calls ``pixel_to_pixel`` in chunks so that we don't blow up the memory with large e.g. GWCS transforms (as mentioned in https://github.com/astropy/reproject/issues/534)
* It improves the call sites of ``pixel_to_pixel_chunked`` (the new name for the function) to avoid excessive temporary memory usage when constructing the inputs/outputs to these functions.

This can significantly reduce memory usage (factor of several to factors of 10x) with, I hope, no downsides.


astrofrog:
pre-commit.ci autofix

PR #625, opened 2026-07-20
Allow Path in reproject_from_healpix

AlecThomson:
The docs of `reproject_from_healpix` state:

```python
def reproject_from_healpix(
    input_data, output_projection, shape_out=None, hdu_in=1, order="bilinear", nested=None, field=0
):
    """
    Reproject data from a HEALPIX projection to a standard projection.

    Parameters
    ----------
    input_data : object
        The input data to reproject. This can be:

            * The name of a HEALPIX FITS file
            * A `~astropy.io.fits.TableHDU` or `~astropy.io.fits.BinTableHDU`
              instance
            * A tuple where the first element is a `~numpy.ndarray` and the
              second element is a `~astropy.coordinates.BaseCoordinateFrame`
              instance or a string alias for a coordinate frame.
```
However, in the current first branch is only checked in 'The name of a HEALPIX FITS file' is a `str` type. This is a simple PR to fix the type checking for `input_data`.

### DIFF
--- a/reproject/healpix/_utils.py
+++ b/reproject/healpix/_utils.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 import numpy as np
 from astropy.coordinates import (
     ICRS,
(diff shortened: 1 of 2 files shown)


reproject/healpix/_utils.py
```diff
@@ -40,14 +42,12 @@ def parse_input_healpix_data(input_data, field=0, hdu_in=None, nested=None):
...
-    elif isinstance(input_data, str):
-        # NOTE: hdu is not closed here.
-        hdu = fits.open(input_data)[hdu_in or 1]
-        return parse_input_healpix_data(hdu, field=field)
-    elif isinstance(input_data, tuple) and isinstance(input_data[0], np.ndarray):
+        return array_in, coordinate_system_in, nested
+    if isinstance(input_data, str | Path):
+        with fits.open(input_data) as hdulist:
+            return parse_input_healpix_data(hdulist[hdu_in or 1], field=field)
+    if isinstance(input_data, tuple) and isinstance(input_data[0], np.ndarray):
```
astrofrog, line 49: Just curious, why was this changed from elif to if? (and same above)

AlecThomson: This was a tiny bit of refactoring on my part. The previous state was very close to a series of guard clauses, so I formulated it more directly that way. I think the logic remains identical. Following a guard clause style, the early returns mean we don't have to worry about the `else` portions from each previous check.

AlecThomson: Something that did just catch my attention now is that the final type error isn't really complete, and it doesn't really tell the caller what ended up there. I'd be tempted to make it something like:

```python
msg = (
    "input_data should be an HDU object, a filename, or a tuple "
    f"of (array, frame), got {type(input_data)=}"
)
raise TypeError(msg)
```

AlecThomson: FWIW - there's even a Ruff rule that enforces this
https://docs.astral.sh/ruff/rules/superfluous-else-return/

Which is maybe why I've picked up the habit!

astrofrog: Feel free to adjust the error message - your suggestion would be an improvement

AlecThomson: Done! Sorry for the slow turnaround

